PR #11, opened 2015-06-04
use angular material instead of bootstrap

championswimmer:


championswimmer:
@creativepsyco should I switch over to material then (and hence merge this PR) ?

creativepsyco:
@championswimmer upto you. I think material is fine. But as long as the templating system is independent I think we can go for anything. Also do squash your commits into 1 before merging. @mananwason is already doing the same.

championswimmer:
Cool !

PR #37, opened 2015-07-10
factor for an extra /event url suffix in testapi

championswimmer:

### DIFF
--- a/app/components/common/ApiJsonFactory.js
+++ b/app/components/common/ApiJsonFactory.js
@@ -5,14 +5,18 @@
 var commonFactories = angular.module('openevent');
 
 commonFactories.factory('ApiJsonFactory', ['$q', '$http', function ($q, $http) {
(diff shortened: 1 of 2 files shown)


championswimmer:
config is defined from config.js file

app/components/common/ApiJsonFactory.js
```diff
@@ -5,14 +5,18 @@
...
-                + 'event/' + config.eventId + '/';
+var baseUrl = (config.use_testApi?'testapi/':config.apiBaseGetUrl) +
+    'event/' + config.eventId;
 
         return {
             getJson: function ($apiEndpoint) {
-                var deferred = $q.defer(),
-                    httpPromise = $http.get(baseUrl + $apiEndpoint);
-
+                if (($apiEndpoint === 'event') && (!config.use_testApi)) {
```
creativepsyco: @championswimmer could make the use_testApi in camel case? then we can get rid of this error.

PR #64, opened 2015-08-19
add Loaders

championswimmer:


creativepsyco:
Are you paging to retrieve the results from the API yet?

championswimmer:
No haven't implemented pagination on the client side yet.

creativepsyco:
Ok I believe still need pagination & perhaps checkout service workers etc. which can allow things to be run in the background.

PR #66, opened 2015-08-24
Update testapi

championswimmer:


assets/js/Utils.js
```diff
@@ -31,4 +31,19 @@ DateUtils.DateDiff = {
         }
 };
 
+function SortUtils () {}
+
+SortUtils.sortBy = function(field, reverse, primer){
+
+        var key = primer ?
+            function(x) {return primer(x[field])} :
+            function(x) {return x[field]};
```
championswimmer: This is a ternary operator, and the semicolon is present where it needs to be

Issue #55, opened 2015-08-13
Tracks are not working


mariobehling:
No clicking possible, no information.
http://fossasia.github.io/open-event-webapp/#/tracks

SaptakS:
I would like to fix this bug. I have seen that it is not possible to click and there is no information either. Where do I get the information to be put there and will an accordion be fine?

championswimmer:
An array of sessions is saved in the $sessionStorage.sessions object. We
need to get sessions from there, and filter then according to the track.

SaptakS:
You mean, when I click a track, I need to show only the sessions corresponding to the track?? Something like that??

championswimmer:
Yes exactly.

note: for later:
Also I want to revisit the way we save the data. Maybe it is not the most
efficient, but yes for now, showing the sessiosn that have the given track
X in it's "tracks" field should be shown when we click on a track

SaptakS:
Ok.. I will send a pull request after I do it. I will ask again if I face any problem.

SaptakS:
I have one more doubt. As I see, the sessions have a title, a type, language and a speaker.. I don't see anything related to tracks. So what exactly do I need to match with the track? Should I add a category all together to the sessions which I can compare with the tracks?

leto:
Howdy,

The API JSON response data for a session should contain some kind of
session identifier which tells the track that the session is a part of. If
not, it should.

Hope that helps,

Duke

On Sat, Sep 12, 2015 at 9:06 PM, Saptak Sengupta <notifications@github.com>
wrote:

> I have one more doubt. As I see, the sessions have a title, a type,
> language and a speaker.. I don't see anything related to tracks. So what
> exactly do I need to match with the track? Should I add a category all
> together to the sessions which I can compare with the tracks?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fossasia/open-event-webapp/issues/55#issuecomment-139843188>
> .
>



-- 
Jonathan "Duke" Leto <duke@leto.net>
Leto Labs LLC http://letolabs.com
209.691.DUKE http://duke.leto.net
@dukeleto http://linkedin.leto.net https://github.com/leto

championswimmer:
http://openevent.championswimmer.in/api/v1/event/1/sessions/page/2

Yeah currently they are all coming out to be null.

I think you this needs be an issue flagged up on
https://github.com/fossasia/open-event-orga-server

SaptakS:
[...]

PR #75, opened 2015-09-24
Tracks fixed .Pull request for #55

SaptakS:


app/components/tracks/TracksController.js
```diff
@@ -31,6 +33,57 @@ tracksModule.controller('TracksController',
...
+            var count = 0;
+            for(var i = 0; i < tdc.sessions.length; i++) {
+                if(trackThis.id == tdc.sessions[i].track) {
+                    count++;                    
+                }
+                
+            }
+            tdc.count = count;
+            var k = 0;
+            tdc.tsessions = new Array(count);
```
SaptakS: but it is an defined within the for. So there is no conflict.

app/components/tracks/TracksController.js
```diff
@@ -31,6 +33,57 @@ tracksModule.controller('TracksController',
                 }, function (error) {
                     console.error(error);
                 });
+
+            ApiJsonFactory.getJson('sessions')
+                .then(function (response) {
+                    tc.Sessions = response.data.sessions;
+                    $sessionStorage.sessions = tc.Sessions;
+                }, function (error) {
+                    console.error(error);
```
SaptakS: console is a feature of firebug which outputs to the browser console. It doesnt need to be defined

leto: Leaving in ```console``` will cause a syntax error on certain older browsers (such as IE) and should be removed. As a workaround, I suggest defining your own debug function that checks for the existence of ```console```

app/components/tracks/TracksController.js
leto: What is the reasoning behind making an alias of ```this```? ```this.track.id``` seems more readable to me than ```trackThis.id``` . Also, I suggest creating a helper function ```count_tracks(this)``` to make the code more readable

app/components/tracks/TracksController.js
```diff
@@ -31,6 +33,57 @@ tracksModule.controller('TracksController',
...
+            for(var i = 0; i < tdc.sessions.length; i++) {
+                if(trackThis.id == tdc.sessions[i].track) {
+                    count++;                    
+                }
+                
+            }
+            tdc.count = count;
+            var k = 0;
+            tdc.tsessions = new Array(count);
+            for(var i = 0; i < tdc.sessions.length;i++) {
```
leto: Can you think of a way to have one helper function that could be used to count tracks and sessions? That would be nice. Also, I suggest using the object equality operator ```===```

app/components/tracks/TracksController.js
SaptakS: ok

app/components/tracks/TracksController.js
SaptakS: I will see to it

app/components/tracks/TracksController.js
SaptakS: As far as I saw, all the other controller codes also used `console`. So, I stuck with `console` . If this needs to be fixed then I guess it needs to be fixed for all. Then it deserves to be an issue.

championswimmer: I think you can continue using console. All modern browsers support console.

SaptakS: I agree.

leto: I suggest reading http://stackoverflow.com/questions/14086675/which-browsers-support-console-log . Your code will not work on IE9. I do not believe breaking the website for millions of potential users is worth it, when a single if statement can fix the problem. Let's support as many browsers as we reasonably can. As a data point, I am using IE9 right now (the horror!), because I am in the hospital and cannot use my own computer. Github made sure their site still works on IE9 and so should we.

SaptakS: My point is we should make this as a separate issue, because all the controller scripts have to be changed to make the modification because all the scripts are using console

SaptakS:
Please check the changes.

championswimmer:
@Duke, I think Saptak is right. Lets make that a new issue. There are many
places in the code we have to change console.log()

So we can make a blanket change as a new pr

creativepsyco:
@SaptakS Please clean up your commits.There is a merge commit inside. Squash and push.

SaptakS:
Done.

creativepsyco:
Your commit involves changes done in e4a01ff75ab5674e28358dc2957c309a2a73cd73 please remove that.

SaptakS:
Done. Sorry for the mistakes.

creativepsyco:
@SaptakS your commits have 6572e1c77234911e1390540f5373dbfcec6bedef which is essentially a merge that you might have done when you were making your fork fresh. You need to revert that merge, rebase with the parent, add your actual commit d00a9a64d0639e7d9c90f558aa2b6e8fcfab0d33, have a sensible commit message and then push. 

Please see https://yangsu.github.io/pull-request-tutorial/ if you submitting a pull request for the first time.

PR #76, opened 2015-10-02
track dialog box added, sessions of respective track shown

SaptakS:
Tracks are made clickable. On clicking a track, a new dialog box opens which shows all the sessions listed in that track. Solves the issue #55

### DIFF
--- a/app/components/tracks/TracksController.js
+++ b/app/components/tracks/TracksController.js
@@ -13,8 +13,8 @@ tracksModule.config(['$stateProvider', function($stateProvider) {
 }]);
 
 tracksModule.controller('TracksController', 
-	['$sessionStorage', '$rootScope', 'ApiJsonFactory',
-        function($sessionStorage, $rootScope, ApiJsonFactory) {
+	['$mdDialog','$sessionStorage', '$rootScope', 'ApiJsonFactory',
+        function($mdDialog, $sessionStorage, $rootScope, ApiJsonFactory) {
 		var tc = this;
         if ( $sessionStorage.tracks === null ||
             typeof($sessionStorage.tracks) == 'undefined')
@@ -22,6 +22,8 @@ tracksModule.controller('TracksController',
             $sessionStorage.tracks = [];
         }
         tc.Tracks = $sessionStorage.tracks;
+        tc.Sessions = $sessionStorage.sessions;
+        
 
         if (tc.Tracks.length === 0) {
             ApiJsonFactory.getJson('tracks')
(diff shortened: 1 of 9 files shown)


testapi/event/1/sponsors
```diff
@@ -6,7 +6,7 @@
       "logo": "http://upload.wikimedia.org/wikipedia/commons/thumb/3/30/Googlelogo.png/320px-Googlelogo.png"
     },
     {
-      "name": "Mozilla",
+      "name": "Moz",
```
creativepsyco: Is there a reason to change the Sample event file?

testapi/event/1/sessions
```diff
@@ -587,24 +623,28 @@
...
+        }, 
+        {
+          "id": 1345, 
+          "name": "Kirsten Fiedler"
         }
       ], 
       "subtitle": null, 
-      "title": "Die Unsichtbaren \u2013 Migration in Europa", 
-      "track": null
+      "title": "States think we're stupid: Internet censorship around Europe since ACTA",
```
creativepsyco: Why is sample data being modified? May I ask?

app/components/tracks/TracksController.js
```diff
@@ -31,6 +33,65 @@ tracksModule.controller('TracksController',
...
+                    tc.Sessions = response.data.sessions;
+                    $sessionStorage.sessions = tc.Sessions;
+                }, function (error) {
+                    console.error(error);
+                });
         }
 
-}]);
\ No newline at end of file
+        tc.showSession = function(track, event) {
```
creativepsyco: yeah actually how are you defining `singleTrack` ? You should consider passing it as a dependency to the controller instead of relying on the global scope of things.

creativepsyco: I think this should help: http://stackoverflow.com/questions/12008908/angularjs-how-can-i-pass-variables-between-controllers

creativepsyco:
Looks okay but I need to know why the sample data is being changed? and if possible can you fix the `singleTrack` variable sharing between the 2 controllers using dependency injection.

testapi/event/1/sessions
SaptakS: Because previously in the sample data all the tracks were empty. But now the API has been modified so the data received is also different. So I changed the test/sample data. Previously, all tracks were null since no value of track was being provided by API. Now that has been fixed.
https://github.com/fossasia/open-event-orga-server/issues/117

SaptakS:
Sir, that can be done but that is a problem for all the controllers and their respective dialog controller. So, if needed to fix, it should be fixed for all the controllers I guess. In that case, shouldn't it be a separate pull request?

creativepsyco:
@SaptakS At least we should not repeat the same mistakes, do you agree? If we can fix it for the current code that you submit, we should. I believe it is worth it.

SaptakS:
Ok. I will see to it.

SaptakS:
I have added dependency injection. Please check.

creativepsyco:
Hi @SaptakS I am sorry this review is taking a bit longer than expected. I am curious to know why did you change the sample data? After this we can get it merged.

SaptakS:
I already mentioned above that the sample data needed to be changed because there was a bug in the API. It always returned null for the track field. So after they fixed it, I had to change the sample data because previously in the sample data, all the tracks were null. So, I used this sample data: http://openevent.championswimmer.in/api/v1/event/1/sessions

creativepsyco:
:+1: Thanks for the info. Merged.

SaptakS:
Thanks

PR #77, opened 2015-10-06
Speaker dialog module shows sessions addressed by the speaker

SaptakS:
The speaker dialog box shows the sessions addressed by the speaker in card layout. Clicking on this sessions opens the dialog module for the particular session

### DIFF
--- a/app/components/speakers/SpeakersController.js
+++ b/app/components/speakers/SpeakersController.js
@@ -88,31 +88,23 @@ speakersModule.controller('SpeakerDialogController',
         }
 
         sdc.count = function(speaker, sessions){
-            var count = 0;
-            for(var i = 0; i < sessions.length; i++) {
-                if(sessions[i].speakers.length > 0) {
-                    for(var j = 0; j < sessions[i].speakers.length; j++) {
-                        if(speaker.id == sessions[i].speakers[j].id)
-                            count++;
-                    }
-                }                    
-            }
-
+            var count = sessions.filter(function(session){
+                return session.speakers.filter(function(sp){
+                          return sp.id == speaker.id;
+                 })
+                 .length > 0;                        
+            }).length;
             return count;
         };
         
         sdc.sessionsDetail = function(speaker, sessions) {
-            var k = 0;
-            var count = sdc.count(speaker,sessions);
-            var spSessions = new Array(count);
-            for(var i = 0; i < sessions.length;i++) {
-                if(sessions[i].speakers.length > 0) {
-                    for(var j = 0; j < sessions[i].speakers.length; j++) {
-                        if(speaker.id == sessions[i].speakers[j].id)
-                            spSessions[k++] = sessions[i];
-                    }
-                }
-            }
+
+            var spSessions = sessions.filter(function(session){
+                 return session.speakers.filter(function(sp){
+                          return sp.id == speaker.id;
+                 })
+                 .length > 0;                
+            });
 
             return spSessions;                
         };
--- a/app/components/speakers/SpeakersController.js
+++ b/app/components/speakers/SpeakersController.js
@@ -40,10 +40,19 @@ speakersModule.controller('SpeakersController',
                 }, function (error) {
                     console.error(error);
                 });
+            ApiJsonFactory.getJson('sessions')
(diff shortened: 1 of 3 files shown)


app/components/speakers/SpeakersController.js
```diff
@@ -77,6 +87,48 @@ speakersModule.controller('SpeakerDialogController',
                 {field: 'Position', value: sdc.speaker.position});
         }
 
+        sdc.count = function(speaker, sessions){
+            var count = 0;
+            for(var i = 0; i < sessions.length; i++) {
+                if(sessions[i].speakers.length > 0) {
```
creativepsyco: should probably guard against speakers being `null`, just to be safe.

SaptakS: OK. I didn't consider that. Will put that check.

app/components/speakers/SpeakersController.js
```diff
@@ -77,6 +87,48 @@ speakersModule.controller('SpeakerDialogController',
...
+                    for(var j = 0; j < sessions[i].speakers.length; j++) {
+                        if(speaker.id == sessions[i].speakers[j].id)
+                            count++;
```
creativepsyco: This function can be replaced by [Array:filter][1] 

```js
sessions.filter(function(session){
     return session.speakers.filter(function(sp){
              return sp.id == 291;
     })
    .length > 0;
})
```
e.g. to search for all sessions by speaker id 291. More concise and clear.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

app/components/speakers/SpeakersController.js
```diff
@@ -77,6 +87,48 @@ speakersModule.controller('SpeakerDialogController',
...
+        
+        sdc.sessionsDetail = function(speaker, sessions) {
+            var k = 0;
+            var count = sdc.count(speaker,sessions);
+            var spSessions = new Array(count);
+            for(var i = 0; i < sessions.length;i++) {
+                if(sessions[i].speakers.length > 0) {
+                    for(var j = 0; j < sessions[i].speakers.length; j++) {
+                        if(speaker.id == sessions[i].speakers[j].id)
+                            spSessions[k++] = sessions[i];
```
creativepsyco: can use the filter method here.

SaptakS:
Sir, I have amended the code to use array filter. Please review.

app/components/speakers/SpeakersController.js
```diff
@@ -77,6 +87,43 @@ speakersModule.controller('SpeakerDialogController',
...
         }
 
+        sdc.count = function(speaker, sessions){
+            var count = 0;
+            sessions.filter(function(session){
+                if (session.speakers.filter(function(sp){
+                          return sp.id == speaker.id;
+                 })
+                 .length > 0)
+                    count++;
```
creativepsyco: The filter function should not be used to do calculations inside. I suppose you are interested in counting the number of sessions by the particular speaker. Then you can just use:

```js
var matchingSessions = sessions.filter(function(session){return session.speakers.filter(function(sp){return sp.id == speaker.id}).length > 0;})

size = matchingSessions.length
```

You can always assume that you will get array back, the filter function must return `true` or `false` and should not do any calculation inside.

app/components/speakers/SpeakersController.js
```diff
@@ -77,6 +87,43 @@ speakersModule.controller('SpeakerDialogController',
                 {field: 'Position', value: sdc.speaker.position});
         }
 
+        sdc.count = function(speaker, sessions){
+            var count = 0;
+            sessions.filter(function(session){
```
creativepsyco: missing return statement here.

creativepsyco:
@SaptakS see my comments above.

app/components/speakers/SpeakersController.js
SaptakS: Ok sir.

SaptakS:
Amended the changes. Please review.

creativepsyco:
:+1: Looks good thanks for the fix.

SaptakS:
(y)

Issue #83, opened 2015-12-09
Unnecessary Margins in Sessions/Speakers details


MaxBorsch:
<b>Summary</b>
When you navigate to the Sessions or Speakers page and then select one of the listed speakers/sessions, extra details will pop up. The details view has an extra empty h2 tag that adds unnecessary margins above and below itself. I noticed that the Track details also has this, however the h2 tag in the Track details has text in it, so the margins look OK. I have included images showing this.

<b>Steps to Reproduct Bug:</b>
1. Start server locally, navigate to web app page. (http://localhost:8080/)
2. Click on Speakers tab or click on Sessions tab.
3. Select a Speaker or Session from the list.
4. The details view will show, in the details view there should not be an extra empty h2 tag that adds unnessesary margins. 

The empty h2 tag is shown here:
![unnessesary element](https://cloud.githubusercontent.com/assets/2924585/11674603/793777e4-9df0-11e5-9861-5dbcc6db238f.png)

Shows unnecessary margins:
![speakers bad](https://cloud.githubusercontent.com/assets/2924585/11674604/7939c350-9df0-11e5-9987-f492de44b0e3.png)

Track details, which has a title so margins look good:
![tracks good](https://cloud.githubusercontent.com/assets/2924585/11674270/2149b630-9ded-11e5-9b85-5ce1fdce29c6.png)

aayusharora:
Hi @MaxBorsch I want to work on this bug. Can you assign it to me ?

MaxBorsch:
Ok, how do you assign an issue? (I¹m not very familiar with GitHub)

From:  aayusharora <notifications@github.com>
Reply-To:  fossasia/open-event-webapp
<reply+002ca029781f652778424c384a841f1bff911fdb2664d99c92cf000000011284242d9
2a169ce073889ef@reply.github.com>
Date:  Saturday, December 12, 2015 at 1:06 PM
To:  fossasia/open-event-webapp <open-event-webapp@noreply.github.com>
Cc:  Max Borsch <maximborsch@gmail.com>
Subject:  Re: [open-event-webapp] Unnecessary Margins in Sessions/Speakers
details (#83)


Hi @MaxBorsch <https://github.com/MaxBorsch>  I want to work on this bug.
Can you assign it to me ?


Reply to this email directly or view it on GitHub
<https://github.com/fossasia/open-event-webapp/issues/83#issuecomment-164172
809> .

nikunjness:
[...]

PR #86, opened 2015-12-12
Unnecessary Margins in Sessions/Speakers details

PiotrKowalski:
I didn't want destroy anything so I created two new things in .css and set one of them to !important 
If you want I can look for these two things and change it in original form.
https://github.com/fossasia/open-event-webapp/issues/83

### DIFF
--- a/app/components/sessions/sessiondialog.html
+++ b/app/components/sessions/sessiondialog.html
@@ -12,7 +12,7 @@
             </div>
         </md-toolbar>
         <md-dialog-content>
-            <div>
+            <div class="sub_content">
                 <md-subheader>{{sdc.session.subtitle}}</md-subheader>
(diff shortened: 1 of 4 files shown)


PiotrKowalski:
Indeed, you think about it? http://scr.hu/3e1h/ryii6
I will repair this. Give me a sec.

PiotrKowalski:
Fixed my mistakes. I think that it is better now.

PiotrKowalski:
Of course:
http://scr.hu/3e1h/voyyx
http://scr.hu/3e1h/d9a6w
http://scr.hu/3e1h/k58nr

MaxBorsch:
Theres also a subtitle under the title that just is empty usually. Would it be visible? Also I think I fixed it with a better way.

PR #92, opened 2015-12-15
Web Manifest + icons

MaxBorsch:
Added Web Manifest and some basic icons + favicon


championswimmer:
Thanks for the PR @MaxBorsch 

If there is a corresponding GCI instance that has not been approved, please link it here in the comments and someone from FOSSASIA will approve it

PR #99, opened 2015-12-27
Add placeholder logo to title bar

darrylyeo:
Adds a placeholder logo to title bar for event organizers to replace with actual event logo.


championswimmer:
Can you squash these two commits into one ? Would be awesome if you could do that :)

darrylyeo:
No problem, done!

Issue #107, opened 2016-02-04
Day header is overlapping the slider


vjs3:
In the sessions tab of the web app header named Day 2 is overlapping the smaller slider.I have seen this issue in google chrome (in linux).
I am attaching screenshot for more clear understanding of the issue.
![openevent - google chrome_002](https://cloud.githubusercontent.com/assets/11405622/12818478/0e7d2452-cb7d-11e5-9f71-0834a678b9a6.png)

tarun14110:
I think this is intentionally done. So that a user can scroll without caring to remember the day. At any point of time he can have a look over above and get an idea, where he is.

vjs3:
Yes I know but my issue is different, it says that day header is
overlapping the scrollbar itself which should not happen.

aayusharora:
@vjs3 ,I got your issue.Working on it.

aayusharora:
fixed here ,#108

rajansavaliya:
Please tell me how to open .js file. I am opening it in chrome and it is not displaying as you put photo. It is just displaying code

PR #108, opened 2016-02-11
#107 Day header is overlapping the slider

aayusharora:


app/components/sessions/sessions.html
```diff
@@ -13,7 +13,7 @@
 
         <md-content style="height: 720px">
             <section ng-repeat="day in sc.Days" ng-if="day.sessions.length > 0">
-                <md-subheader class="md-primary"> {{ day.label }}</md-subheader>
+                <md-subheader style="margin-right:35px;" class="md-primary"> {{ day.label }}</md-subheader>
```
chenzihui: I don't think its a good idea to inline the CSS. It's probably better if this is done in CSS / SCSS files.

aayusharora:
@chenzihui ,You are correct,the reason behind was that the styling of the whole document is inline,I can create a seperate css for the whole document,if it is required.

creativepsyco:
@aayusharora please see https://github.com/fossasia/open-event-webapp/blob/master/assets/scss/main.scss

Issue #106, opened 2016-01-24
Accept rating for talk.


creativepsyco:
Allow user to rate a session and provide feedback.

Will require some sort of authentication, fb, google, twitter etc.

SaptakS:
I would like to work on this. I guess the authentication part will be done using the open-server repo if I am not wrong.

AmayAgrawal:
I think that we can first log in the user using any ine of the above option.Is authentication be done here or the open-server.

creativepsyco:
you can login the person through any means on the client side, be it
google-auth, or github/twitter etc. but the server should only care about
the email address, that's unique enough for the identification of the
review/rating.

>

AmayAgrawal:
I have almost done this .I have added a functionality which authenticates user using google and redirects to new page where the user can rate the individual talk and provide feedback.After this what I have to do.I have to directly put into database or what this feedback and rating??

creativepsyco:
Here is a diagram of how all this fits together

The server does not care how you authenticate the user, it is just interested in the session review and the email of the user which is submitted via the client only after the user has authenticated through some means at the client side. It can be github, google, or any other form of oauth 2.0 relevant signin. I am sure there are packages and open source libraries which make this process easier. At the end the client should store the email through some persistent mechanism, cookies, filestorage, localstorage etc.

![server architecture](https://cloud.githubusercontent.com/assets/547235/13099620/132e1b86-d56f-11e5-8888-dd789e07ad3f.png)

aayusharora:
@SaptakS ,@AmayAgarwal,How much is left with this issue,I have already in middle for the client side part of it ,the rating part .Is it done or should I do it ?

AmayAgrawal:
I did it .The PR is not  been verified yet but.

PR #119, opened 2016-02-24
Social sharing buttons

AmayAgrawal:
I have at present added a functionality to share it on fb,twitter and google.I have to make the following changes in index1.html also which is pending to get merged for issue #106 .After adding this sharing button it looks something like this.If needed any change please let me know.The sharing buttons are in left side navigation bar.
screenshot from 2016-02-24 22 53 57


championswimmer:
does not go well with the material theme

Issue #146, opened 2016-03-20
Report for Errors in Twitter Wall


aayusharora:
After the merge , some errors are there in the section 

-  Unknown Provider for the Tweet directive
-  Loader has a wrong path 
-  XMLHttpRequest cannot load http://loklak.org/api/search.json?q=fossasia. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://fossasia.github.io' is therefore not allowed access. The response had HTTP status code 503.

All these are not there when working locally , except XmlHttpRequest.

aayusharora:
![1](https://cloud.githubusercontent.com/assets/12194719/13906828/66a90b1c-ef06-11e5-9c4d-98bc9e70973f.png)

aayusharora:
@championswimmer ,solved " Unknown Provider for the Tweet directive ", Here #147

aayusharora:
@championswimmer ,solved Loader has a wrong path Here #148

aayusharora:
@championswimmer ,solved Cross Domain Problem Here  #149

aayusharora:
All done , Closing this

PR #147, opened 2016-03-20
#146 Injector error solved

aayusharora:
Source for the solution : https://docs.angularjs.org/error/$injector/unpr?p0=eProvider%20%3C-%20e%20%3C-%20tweetDirective


championswimmer:
Still doesn't solve the cross domain problem

aayusharora:
@championswimmer ,working for the same :)

PR #150, opened 2016-03-23
sponsorsLogoMadeVisible

SilentFlame:
due to the presence of the class name="sponsor-logo" it was making the view of the sponsor image somehow 0x0 pixels.
sponsors

changing the class name makes it work all fine.
final_sponsor
initially there were some `#shadow.root` style which was making all this, but now unable to see that style attribute.

@championswimmer can you explain it please.

### DIFF
--- a/app/components/sponsors/sponsors.html
+++ b/app/components/sponsors/sponsors.html
@@ -19,7 +19,7 @@
                           md-rowspan="{{2}}"
                           md-colspan="{{2}}"
                           md-colspan-sm="1">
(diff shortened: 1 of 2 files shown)


aayusharora:
@SilentFlame , I am not finding any new things in the commit . Kindly explain.
Class name changed from sponsor-logo to logo ?

SilentFlame:
@aayusharora problem is that initially when it was `sponsor-logo` it have some style files somewhere,
which is

PR #131, opened 2016-03-08
right navigation bar made proper.

SilentFlame:
The left navigation bar made consistent throughout the pages `session, speakers, tracks`.
@mariobehling , @jig08 
have a review if works as expected, then can merge and close Issue.

PR #165, opened 2016-04-08
consistency in side-bar

### LINKED ISSUE #132: scrolling behaves differently on speakers page.
Sometimes it works something similar to sessions like:
right

but sometimes doesn't like :
wrong

this all happens in alternate clicks on speakers tab on left.
also the top header of search by name and org should stay static while scrolling down.

SilentFlame:
@championswimmer 
replica of #131 and resolves #132. just improved the `speaker.html` sections' side-bar.
since tracks view has been changed so need to first make it like before.so will be working on that.


championswimmer:
cool ! looks good

PR #168, opened 2016-04-12
decentring Track and Map

SilentFlame:
@championswimmer 
it's made as it was before, have a look.

now the sponsors section also works fine.

**previously**
map
sponsors
tracks

**After change**
map_final
sponsors_final
track_final

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,7 +8,7 @@
 
     <!-- inject:css -->
     <!-- endinject -->
-    
+
     <meta charset="UTF-8">
     <meta name="viewport" content="initial-scale=1" />
     <title > {{ appCtrl.appTitle }} </title>
@@ -32,13 +32,13 @@
         </md-sidenav>
 
         <md-content flex layout-padding>
-            <div layout="column" layout-fill layout-align="top center">
+            <div layout="column">
                 <div ui-view></div>
             </div>
             <div flex></div>
         </md-content>
     </div>
-    
+
 
     <!-- inject:js -->
     <!-- endinject -->


aayusharora:
@SilentFlame , it is already as it is on my local (not centered). Even all seems perfect at http://fossasia.github.io/open-event-webapp/#/tracks , http://fossasia.github.io/open-event-webapp/#/sponsors , http://fossasia.github.io/open-event-webapp/#/map, what is the error?

SilentFlame:
@aayusharora when you sync the upstream, due to the update in the structure and Angular from `angular: 1.4.4` to `angular: 1.5.3` this kind of changes happen.

and since the [Test Version](http://fossasia.github.io/open-event-webapp/#/sessions) is the `gh-pages` branch not the `master` which we are changing hence the changes are not reflected on `gh-pages`

aayusharora:
@SilentFlame ,I completely agree , but things are not working as you said.
 I am already sync with master , still there is no problem I am facing.
 Please see this even the class** layout-fill layout-align="center"** is there , but the layout is not centered still (means its correct without removing the class you have removed in the commit to decenter)
Snapshot :

championswimmer:
@SilentFlame @aayusharora what browsers are you on respectively?

aayusharora:
@championswimmer  , I am using Google Chrome Version 48.0.2564.103 (64-bit)

championswimmer:
I can see decentering on Firefox

championswimmer:
As in, centered column, with paddings on left and right. On the current
master head. Have not tried this PR yet.

SilentFlame:
@championswimmer  @aayusharora I'm tested the code in both `Google-chrome` and `firefox`.

same happening in both the browsers in my case.
**initialy**
track
**Finally**
track_new

aayusharora:
@SilentFlame , "layout-align-start-center" is the class that is doing all the stuff. It is not present in my code nor on the remote , please check your index.html with https://github.com/fossasia/open-event-webapp/blob/master/src/index.html.
Your code

Original Code

SilentFlame:
@aayusharora I accept the Original code is as below
original

but now since it seems the `Trial Version` is updated it too reflects the centring of the `tracks` and `map` section.
tracks

I'm trying to find out the reason why it creates a class with with such name which are present in that div.

PR #174, opened 2016-04-15
new package.json

aayusharora:
Updated these two dependencies:
Screenshot:


juslee:
@aayusharora update npm-shrinkwrap after updating packages.

aayusharora:
@juslee ,I have read the [docs ](https://docs.npmjs.com/cli/shrinkwrap) and applying the same .
I am getting the following error , any idea will help.

aayusharora:
@juslee , its done , Do we have to require dev dependencies also?

championswimmer:
The way the app is structured, npm dependencies are required to 'build' the
project, and bower dependencies is required to 'run' it

That is why there *are no* dev-dependencies at all. All npm dependencies
are in a way dev-dependencies.

championswimmer:
@juslee, the way I had set the project up, there should be no 'prod'
dependency in package.json, only 'dev' dependencies.

All runtime dependencies are part of bower.json

PR #179, opened 2016-04-17
Update .jshintrc

aayusharora:


aayusharora:
After writing predef , I found only one post http://jnst.hateblo.jp/entry/2013/12/22/JSHint_%E3%81%AE_predef_%E3%81%AF_deprecated indicating predef is deprecated and we should global. The post is another language , have translated :),
Hope it should work.

Issue #184, opened 2016-04-26
chore: enable unit testing


juslee:
Look into using mocha or some unit testing framework and start the initial unit tests. This is to enable code coverage later.

PR #185, opened 2016-04-26
#184 Karma jasmine added package.json

aayusharora:


aayusharora:
Will generate the new one , after figuring out shrinkwrap warnings . :)

PR #220, opened 2016-05-31
Speaker View

aayusharora:
@juslee , @championswimmer  
**LIVE VERSION** http://peddler-maurice-60217.bitballoon.com/#/speakers
Its ready to merge .

### DIFF
--- a/index.html
+++ b/index.html
@@ -45,6 +45,7 @@
     <script type="text/javascript" src="./src/appComponents/sidenav/SidenavController.js"></script>
     <script type="text/javascript" src="./src/appComponents/sessions/SessionsController.js"></script>
     <script type="text/javascript" src="./src/appComponents/speakers/SpeakersController.js"></script>
+    <script type="text/javascript" src="./src/appComponents/speakers/Speakersdirective.js"></script>
     <script type="text/javascript" src="./src/appComponents/map/MapController.js"></script>
     <script type="text/javascript" src="./src/appComponents/sponsors/SponsorsController.js"></script>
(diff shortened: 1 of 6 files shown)


mariobehling:
I am confused - why are you merging this into GSoC 2015? Instead of 2016?

aayusharora:
@mariobehling , As I explained earlier , all the configs in GSOC16 created by @championswimmer , are based on Angular2.0 . We have set the later milestone to update it to Angular2 , as we also are not familiar with typescript , and needs time to learn it . In GSOC15 branch we will be ready with our views and functionalities with Angular 1.5 .

aayusharora:
@juslee , solved issue #221

aayusharora:
@juslee ,@championswimmer , it also solved issue #222

mariobehling:
@aayusharora It looks like you are simply continuing the UI of last year. It is slow and search did not work. I suggest we clean up and start with a basic thing, that works. I think I am repeating myself now on many different occasions.

aayusharora:
@mariobehling ,
1.  I was reusing the code like services . But , the modules say like Speakers , Sessions were strongly binded together in last year code , Angular Material was also making the things slow .
2. In this PR , All the code of speaker have become separate , there is very less coupling .
3. The use of directives for making components is always advantageous, Unlike previous approach , In this PR the searchbox and the Cards are separate directives , which can be used anywhere .
4 . The problem which I have seen with the last year UI , was the unnecessary use of space on the page 
     Here each row contains three speakers information which is a better design approach 
I will be more clear after tomorrows meeting .

PR #247, opened 2016-06-08
config files added

aayusharora:
@championswimmer @juslee , It will solve es6 config issues 
Added .hound.yml
Added .eslintrc


championswimmer:
I already had added eslintrc though.

PR #255, opened 2016-06-09
Rewrite schedule.css to application.scss

aayusharora:


aayusharora:
@juslee @championswimmer  @mariobehling  , it can be merged .

src/generator/backend/_scss/application.scss
```diff
@@ -149,61 +150,76 @@ body {
...
 
-.session-link > .fa:hover {
-	color: #2196F3;
-}
+  > .fa {
+    font-size: 18px;
+    color: #757575;
+
+      &:hover {
+        color: #2196F3;
```
championswimmer: @aayusharora 
I think you should un-hardcode these color definitions.

aayusharora: @championswimmer  , I think the same , doing that . But what should be the solution for values like $code-color . The $code-color is set as white in config , but overwritten in light and dark themes . If I replace all hardcoded white everywhere with $code-color , it will change to gray in all the places in light theme which we don't require .

PR #270, opened 2016-06-11
Readme added

aayusharora:


championswimmer:
If a PR is about README, then let there only be README related changes in it 

Make other changes separately

aayusharora:
@championswimmer  , do we need this now #274

championswimmer:
@aayusharora You can add the description of the project and what are the purpose of each folder etc in this PR. The current README I added does not have those things.

PR #277, opened 2016-06-12
Added Component 1 tracks

aayusharora:


championswimmer:
One thing here. Instead of making a async.series, you can put it in the same WRITING block. 

Writing the schedule.html and tracks.html can run asynchronously parallelly. Not necessary to do serially one after another.

aayusharora:
@championswimmer , Can I change it next PR , generating the next PR in a minute with markup. Merge this one 
:)

championswimmer:
Cool. merging this. 
Do the markup in the next PR, and also put all the writing tasks in the same async.series block. 
You can use a async.parallel driver for it.

Issue #311, opened 2016-06-18
Error page for handling wrong JSONS or endpoint


aayusharora:
Use Cases
 A user may upload 
    1. wrong file format.
    2. incomplete files.
    3. Wrong end point 

In all these cases or similar cases. a frontend Error page should handle this and provide a button to go to the index page again.

aayusharora:
Done !!

PR #319, opened 2016-06-20
Errorpage

aayusharora:
Issue Related #311


championswimmer:
FUTURE: We need to add more information in the error page. If it's wrong JSONs, then user should know that he uploaded malformed JSONs.

PR #363, opened 2016-06-25
make call to getJsonData only once

championswimmer:
redundant calls = more memory. more cpu cycles
bad bad bad


championswimmer:
@aayusharora please review

PR #409, opened 2016-07-04
Download json api

championswimmer:


championswimmer:
@aayusharora this is done now, but our event.json is not same as orga-server sample v2 API.
So it works only with this endpoint 
https://raw.githubusercontent.com/fossasia/open-event-webapp/master/mockjson/
but not with this endpoint for eg
https://open-event.herokuapp.com/api/v2/events/4

PR #456, opened 2016-07-11
Dark theme removed

aayusharora:


championswimmer:
let the scss remain, just remove the option from the frontend it becomes disabled.

championswimmer:
Done via my PR.

PR #461, opened 2016-07-11
Improve formatting and cache variable for popover.js

aakash-cr7:
Instead of calling jQuery 3 times to search `$('.pop-box')` we can cache the `.pop-box` at the top and use it later on anytime. This would be more efficient as well as a good practice. What do you guys think?

Also improved formatting.

### DIFF
--- a/src/generator/backend/assets/js/popover.js
+++ b/src/generator/backend/assets/js/popover.js
@@ -1,7 +1,8 @@
 /* Created by aayusharora on June 20, 2017 */
 
 $(document).ready(function() {
-  $('.pop-box').hide();
+  var popbox = $('.pop-box');
+  popbox.hide();
   if($(window).width() < 768) {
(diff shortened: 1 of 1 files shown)


src/generator/backend/assets/js/popover.js
```diff
@@ -1,7 +1,8 @@
 /* Created by aayusharora on June 20, 2017 */
 
 $(document).ready(function() {
-  $('.pop-box').hide();
+  var $popbox = $('.pop-box');
```
championswimmer: just call is `popbox` and not `$popbox` then ? 
$() is for jQuery stuff only. If it's our on var, just name it simple in letters.

aakash-cr7:
@championswimmer removed the '$' sign from the variable.

championswimmer:
Thank you @aakash-cr7 :)

PR #470, opened 2016-07-12
Add compression middleware

championswimmer:

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "archiver": "^1.0.0",
     "async": "^2.0.0-rc.6",
     "chai": "^3.5.0",
+    "compression": "^1.6.2",
     "connect-domain": "^0.5.0",
     "express": "^4.13.4",
     "fs": "0.0.2",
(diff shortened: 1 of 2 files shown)


championswimmer:
@aayusharora Please check out differences between

https://opev-webgen-dev.herokuapp.com/ (this PR)
and 
http://open-event-generator.herokuapp.com/ (master)

aayusharora:
@championswimmer Can you please see this before I compare 

This comes here https://opev-webgen-dev.herokuapp.com/

aayusharora:
@championswimmer  it does not work for me. We have to figure out a solution as early as possible. This is taking 13 min to finish loading. The data will increase and hence the loading time. The only drawback here is we can't load data with scrolling. Loading all the data at once is the major problem of the page
Here is the snapshot.Please give suggestions.
fail

aayusharora:
@championswimmer Though an improvement is there but that is not sufficient to work
This PR


Master

championswimmer:
As per my tests across 3 browsers, it is decreasing load time by 20%
So I am merging this. We will have to look at even more methods to increase performance

PR #499, opened 2016-07-18
move to using socket instead of POST

championswimmer:


championswimmer:
The generator shows status of build now 
gen

Issue #501, opened 2016-07-19
UX Improvement Track page


aayusharora:
- [x] Fix Responsiveness
- [x] Add speaker organisation and social links along with image
- [x] Change popover box to close on un-hover state rather than click as sched.org
- [x] Implement Dynamic Footer
- [x]  UX :Add Hover state to track hovered

aayusharora:
@mariobehling  @championswimmer  Solved with PR #502

mariobehling:
Please provide a screenshot in future.

aayusharora:
Sure @mariobehling , closing this its done #502

Issue #500, opened 2016-07-19
Make Dynamic Footer with popup dissimilar to sched.org


aayusharora:
The footer used by sched.org are not dynamic which causes basic problem for the last item to take its height 
![13](https://cloud.githubusercontent.com/assets/12194719/16956759/4ddd16b4-4df7-11e6-83df-99665a616c9f.png)

This will make a problem as it will not provide the pop-over height to take its space on page.

aayusharora:
@mariobehling @championswimmer done with #502

mariobehling:
@aayusharora It would be good to provide a link from now on to the sample webapp. Could you do that, please?

aayusharora:
Will be providing link ahead. This is merged and good to close.

PR #502, opened 2016-07-19
Dynamic footer and Track UX

aayusharora:
@championswimmer Its done Please check and merge 
- [x] Issue #500  Footer should always remain at bottom of page 
 done here https://drive.google.com/file/d/0BylUgB3l8Lc6bDg4UFVlMnZXeDA/view
- [x] Change popover box to close on un-hover state rather than click as sched.org
- [x] Make pop-over responsive 
- [x] UX :Add Hover state to track hovered
- [x] Add speaker social links

Solve issue #500 #501


championswimmer:
looks nice. merging :)

Issue #504, opened 2016-07-21
Add Track names with their color codes on room page 


aayusharora:
The roompage has colorful session names but it's not written anywhere that which color is from which track. This is confusing for a user to go back to track page and check the room page again.
Please make the information available on room page as sched.org does.

- [x] Add title of track with the color it belongs too.
- [x] Make the portion work well on small screens as sched.org

mariobehling:
Please ensure, you are using the color defined in the tracks.json. You can see an example here https://github.com/fossasia/open-event/blob/master/sample/OTS16/tracks.json

aayusharora:
@mariobehling Yes, definitely all colors are coming from tracks.json.

aayusharora:
Done here #527

PR #510, opened 2016-07-22
merge development to master

championswimmer:
periodic merge into master. 

dev site : http://opev-webgen-dev.herokuapp.com 
If all good, we can merge

PR #514, opened 2016-07-22
Color codes with Title on Room page [ Do not Merge ]

aayusharora:
Issue #504

### DIFF
--- a/src/generator/backend/assets/js/popover.js
+++ b/src/generator/backend/assets/js/popover.js
@@ -1,21 +1,21 @@
 /* Created by aayusharora on June 20, 2017 */
 
-$(document).ready(function() {
-  var widthWindow = $(window).width();
-  var popbox = $('.pop-box');
-  var headerpop = $('.header-pop');
-  var outerContheight= $('.main').offset().top + $('.main').outerHeight()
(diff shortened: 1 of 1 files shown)


championswimmer:
Do not submit comment before image has been uploaded as attachment :p

aayusharora:
@championswimmer  Yeah :p

championswimmer:
@aayusharora does the color code box stay sticky ? or does it scroll away when we scroll down ?

championswimmer:
## hold this merge before #510 is done

aayusharora:
It scroll away .. @championswimmer

championswimmer:
@aayusharora can you make it stick ? 
It's good to have the glossary always available

aayusharora:
@championswimmer I have already given a thought to that. But when we think this for small-screens the sticky track-bar will fail because in small screens the track-bar should come below the tracks above the footer.
This is the reason why sched.org does not keep it sticky. If you can think of an alternate approach please mention.

aayusharora:
@championswimmer Sorry for merging it on my own but that was urgent to check some of the code changes. This was pending for three days and already reviewed by you. Please make a separate issue for sticky track-title if it is needed.

Issue #518, opened 2016-07-23
Fix Scroll bar issues and implement design without scroll bars


mariobehling:
Parent Issue: https://github.com/fossasia/open-event-webapp/issues/515#issuecomment-234709757

In the below screenshot you can see 4 scroll bars. There should only be one scroll bar, the one to scroll from top to down.

* [x] Ensure no scroll bars in the scroll over content
* [x] Ensure the page does not scroll from left to right
* [x] Ensure all is working well on mobile screens

![screenshot from 2016-07-23 12 59 46](https://cloud.githubusercontent.com/assets/1583873/17077601/8b783e46-50d5-11e6-8810-05509db7fc05.png)

aayusharora:
@mariobehling  This will be solved after merging  #529 Please check
![1234](https://cloud.githubusercontent.com/assets/12194719/17191514/d293b1b2-5467-11e6-96e5-b3b75f295ec7.png)
![1236](https://cloud.githubusercontent.com/assets/12194719/17191582/2bf4e2ee-5468-11e6-9aa6-e29c7547e52a.png)

Issue #516, opened 2016-07-23
Test pages, images, links, design, on page


mariobehling:
Before you implement more features, please test the whole site in detail, e.g.
* [x] images are not shown (no need for explanation, rather open an issue and outline a solution)
* [x] Check all links, e.g. home screen not linking to defined home
* [x] Rethink usability, why are there different font sizes in dropdown menu?
* [x] Pages in small resolution are not using full width and breaking style sometimes. Examples below.

Words broken and why not using space on the right?
![screenshot from 2016-07-23 05 43 42](https://cloud.githubusercontent.com/assets/1583873/17075602/bdacda94-5098-11e6-8cd3-3695af8af3b9.png)

Why not full width?
![screenshot from 2016-07-23 05 44 13](https://cloud.githubusercontent.com/assets/1583873/17075603/c1a6db54-5098-11e6-9354-ff4f3299a910.png)

championswimmer:
assigning @aayusharora because he has worked on the styling of the pages mostly.

championswimmer:
@mariobehling wanted a quick clarification. 

The "home" button should link to ? The "event_url" in the json or the sessions page ?

mariobehling:
It depends on how we define it. At the moment it neither directs to the event_url nor the app start directory. The logo correctly links to the event URL. So, I think this is fine and we make the "Home" button direct to the "app root"

championswimmer:
thanks! will do the same

championswimmer:
home page link done here 
https://github.com/fossasia/open-event-webapp/pull/530

aayusharora:
Null images replaced by generic image, Track page style not breaking now solved here #529
![1234](https://cloud.githubusercontent.com/assets/12194719/17193378/60d5117a-5470-11e6-9665-e4b68e3ed255.png)

championswimmer:
Okay, images with links like this 
`photo.jpg?quality=89&w=800` 
are not working correctly

aayusharora:
@championswimmer  It is done and can be closed

aayusharora:
@championswimmer I am closing this, Please open another issue if you still find some images not working.

Issue #515, opened 2016-07-23
Avoid internal page scroll bars


mariobehling:
A number of areas have internal scroll bars, e.g. speakers page and on scroll over session information on tracks page etc.

List all pages with internal scroll bar here and implement a style without scroll bars.

aayusharora:
Hi,@mariobehling
Areas having scrollbars 
1. Pop-overs
2. Speaker section
Internal Scrollbars are not necessity but requirement. For example let's say the the speaker section. We don't want the cards to be of unequal sizes. So we fix the max-height of each card and if the speakers description exceeds the height we give a scrollbar to prevent overflow text.
Hence you can see for better UX
1. No scrollbar is present until you hover the card.
2. Only those cards have scrollbars whose speaker description exceeds maximum card limit. This is intentionally done and will not cause any problem.

Pop-overs
Currently in the design I have set height of pop-over as auto means it can take height according to data but Its necessary to set a maximum height.Obviously No internal scrollbar will come if number of speakers are less. Hence it depends on data just to cover overflow problem.

mariobehling:
Ok, thanks for the listing. Now please implement

* [ ] No scroll bars in 1. Pop-overs
* [ ] Design with no scroll bars in 2. Speaker section (make a screenshot how it will look like.
* [ ] Ensure all works well in mobile

PR #529, opened 2016-07-26
Frontend-issues

aayusharora:
Issue related #516 
- [x] images are not shown (no need for explanation, rather open an issue and outline a solution)
  Issue https://github.com/fossasia/open-event-orga-server/issues/1877#issuecomment-235212941 will   be adding a placeholder for null images.
- [x] Check all links, e.g. home screen not linking to defined home
- [ ] Rethink usability, why are there different font sizes in dropdown menu?
- [x]  Pages in small resolution are not using full width and breaking style sometimes
- [x] No X-axis scroll bar #518 Parent issue #515
- [x]  Scrollbar removed from Pop-over

    Change in UX    
1. Inline sessions to utilize white space.
2. Full width container to contain the elements.


aayusharora:
@mariobehling about the comment
**Rethink usability, why are there different font sizes in dropdown menu?**
Can you please elaborate with a snapshot. I am feeling difficult to find it exactly.

mariobehling:
No, problem. Do first what you can and we follow up step by step.

championswimmer:
@mariobehling @aayusharora can we merge this first ? right now there are UI fixes

The **rethink usability** topic is open ended, we can do that in a future PR. We can discuss that more in a separate issue too if we want

aayusharora:
Yes @championswimmer It's already cover lot of fixes, kindly merge.

PR #536, opened 2016-07-29
Mention Errors on the Front End

abhishek97:
Mention error explicitly on the front-end rather silently logging them
on server side

### DIFF
--- a/src/generator/backend/generator.js
+++ b/src/generator/backend/generator.js
@@ -79,13 +79,14 @@ exports.createDistDir = function(req, socket, callback) {
 
   async.series([
     (done) => {
+        socket.emit('live.process', {status: "Cleaning dist folder"});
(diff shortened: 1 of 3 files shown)


src/generator/backend/generator.js
```diff
@@ -79,13 +79,14 @@ exports.createDistDir = function(req, socket, callback) {
 
   async.series([
     (done) => {
+        socket.emit('live.process', {status: "Cleaning dist folder"});
       distHelper.cleanDist(appFolder, (cleanerr) => {
         console.log('================================CLEANING\n\n\n\n');
-        socket.emit('live.process', {status: "Cleaning dist folder"});
```
championswimmer: Why change the position of this event emit ?

championswimmer: Nvm, that's better.

PR #554, opened 2016-08-01
Sponsor Section Added  

aayusharora:
The problem with the recent Sponsor section is that it does not vary with data. Let's consider some cases
1. When there is only one Image at first and third level ( According to layout) => The image must be centered.
2. If there is no level defined is JSON, all images must be of the same size.
3. The images must be completely responsive.

The current PR checks these cases and automatically adjusts images in each case.
@mariobehling @championswimmer Please check


championswimmer:
LGTM

PR #601, opened 2016-08-08
remove sponsors from tracks page as they are in frontpage now

championswimmer:


championswimmer:
@aayusharora 

i removed it from schedule.hbs, as it is in event.hbs now. also layout was not correct in schedule.hbs

please check and merge

aayusharora:
@championswimmer  Can you please reverify, It seems you have removed them from event.hbs instead.
Seems Dangerous.

championswimmer:
@aayusharora fixed now

PR #614, opened 2016-08-11
fix some issues with images not downloading or copying from zip

championswimmer:
fixes two things - 

 - background image and logo not being copied correctly from zip or not downloading from api
 - urlencode was encoding the '/' character too. that was wrong and was corrupting some images.


championswimmer:
@aayusharora if it does not work, please post the node logs here. I'll take a look. It is working on my end though

PR #630, opened 2016-08-15
Fix background image

championswimmer:
The parameter in the API is background_image not background_url

ALso, to check both null and undefined, we must check `!= null`


championswimmer:
@aayusharora please check merge this

Issue #641, opened 2016-08-16
Fix Sponsor section with more than 3 levels


aayusharora:
After new change in Open-event Fossasia sample. Sponsor section result in bad layout when number of levels increased to five.
Please implement a solution that works in all cases.

aayusharora:
@championswimmer I am working for this.

aayusharora:
Ok so after thinking how can we make this implementation possible so that the sponsor section become flexible with levels.

1. We have to calculate the number of sponsors at each level for this.
     a) **Level 1** => Sponsors should be large in size **Max-height: 120px**
        Case 1: Number of sponsors is 1 and 2
        Implementation: Make it centered in the row/container containing them.
        Case 2: More than 2
        Implementation : In a single row , make four columns each of 25% width to contain them, rest four   will come in next row.

    b) **Level 2* =>  Sponsors should be medium in size **Max-height: 60px;**
          Case 1 and 2 of level1 also applies here. 
    c)  Level 3=> Sponsors should be smaller in size  **Max-height: 55px**
          Case 1 and 2 of level1 also applies here.
     d) No levels or more than 3 level   **Max-height: 55px**
         Sponsors should be exactly same as fossasia.org website. 

Please consider the output from above code 
![levelwise](https://cloud.githubusercontent.com/assets/12194719/17714711/7eb139cc-641e-11e6-91b2-4a0946232ec5.png)

![ots16](https://cloud.githubusercontent.com/assets/12194719/17714727/980386a0-641e-11e6-90fc-07e49bf26e85.png)

mariobehling:
Just keep it as it is at the moment and focus on other tasks to make the application work. Also improve design e.g. 100% width of sponsorship section.

aayusharora:
> Just keep it as it is at the moment

@mariobehling 
It is disturbed at the moment because of levels added in Fossasia JSON. It will not work like that, From my side both of the above-mentioned designs are ready and will not take more than 1 min to generate PR. Please suggest the better one from them.
I am focussing on other tasks as well.

mariobehling:
Ok, just follow the second approach maybe.

PR #645, opened 2016-08-17
Correcting Layout for Sponsor section

aayusharora:
Issue #641 
@mariobehling @championswimmer Please review and merge
Before 

After


championswimmer:
if we are using same sizes for all, can we add a header for each section (Platinum, Gold, like that) of sponsors ? 
It should be easily visible (not tooltips) which tier a sponsor belongs to. They pay money for that only.

aayusharora:
@championswimmer Tooltip is only to avoid confusing UI. We can add names too by removing tooltip later. Please merge this.

PR #646, opened 2016-08-17
upload in background and add progressbar for upload

championswimmer:


championswimmer:
this is ready to be merged now @aayusharora 

please take a look

championswimmer:
Can be tried out here https://opev-webgen-dev-pr-646.herokuapp.com/

aayusharora:
@championswimmer  Just checked it out 
1. Two loaders very confusing ?
2. It animates completely with 70% ( showing all blue) but still button remains disabled?
3. Is there a need to show ETA values ?

mariobehling:
> Can be tried out here https://opev-webgen-dev-pr-646.herokuapp.com/

Not working for me. No progress bar for uploads.

championswimmer:
@mariobehling here it is working for me. on the given link only. 

record3

championswimmer:
> 1. Two loaders very confusing ?
What do you suggest ? hide upload loader when generate starts ? 

> 2. It animates completely with 70% ( showing all blue) but still button remains disabled?
What zip did you use ? what size of zip file ? 

> 3. Is there a need to show ETA values ?
Yes I believe it is good to show that. User's should get an idea

aayusharora:
@championswimmer 
> 1. What do you suggest ? hide upload loader when generate starts ?
Yes, I would suggest hiding the upload loader when upload is done, no need to show that, There should be only one loader ( may be doing all the stuff ).
> 2.What zip did you use ? what size of zip file ?
       It is fossasia sample of nealy 19.8 mb.
> 3. Is there a need to show ETA values ?
 ok we can improve on this later.

PR #658, opened 2016-08-19
fix event variable in firefox

championswimmer:
Signed-off-by: Arnav Gupta <arnav@codingblocks.com>


championswimmer:
@aayusharora please check this. 

on firefox, event variable is not global

mariobehling:
which issue is this?

PR #663, opened 2016-08-20
fix zip uploads for large zips [STILL TESTING] 

championswimmer:
Signed-off-by: Arnav Gupta <arnav@codingblocks.com>


mariobehling:
Let's test together.

Issue #671, opened 2016-08-22
Rooms page bugs: Missing date links, session time confusing, no session info


mariobehling:
The rooms page is still missing basic information including
* [x] no date top links (like in tracks page)
* [x] no expanded session info on click (like in tracks page)
* [x] confusing time format - it should be the same as on tracks page
* [x] Room name Dalton Hall repeating

aayusharora:
@mariobehling @championswimmer Taking this

aayusharora:
Closed via #688

mariobehling:
Thank you, good progress. Now, please ensure the implementation on the page stay within out design and follow the content display of other pages, e.g. the tracks page

* [ ] Ensure in the un-expanded session you show the same information as on the unexpanded version of the tracks pages, e.g. Show: A Funny Thing Happened On My Way To The Science Centre, Harish Pillay (RedHat), Keynote of 20 minutes
* [ ] Ensure you use the same font and color of font as in other pages, including
   * [ ] Time
   * [ ] Text/description/icons etc in expanded section
   * [ ] Align date buttons in the same way as on tracks page

mariobehling:
Please use our gray color set for text.

![screenshot from 2016-08-28 09 07 04](https://cloud.githubusercontent.com/assets/1583873/18032215/ef93b2b2-6cfe-11e6-9efb-5f3ec065cdf8.png)

aayusharora:
Completed this via #719 #722

PR #678, opened 2016-08-22
Removing Repeated Room names

aayusharora:
- [x]  Bug removing repeated Room names ( Venue page ) #671


aayusharora:
@championswimmer Codacy is still unable to analyse it.

PR #679, opened 2016-08-22
Featured speakers only on home screen

championswimmer:
# this will work only after https://github.com/fossasia/open-event/pull/88 is merged


championswimmer:
how it looks when the json has featured = true for some speakers 

screenshot from 2016-08-22 22-45-03

Issue #674, opened 2016-08-22
Make top bar semi-transparent


mariobehling:
Please make the top bar semi-transparent like at the FOSSASIA site.

championswimmer:
it's done

PR #680, opened 2016-08-22
Top bar made semi-transparent

aayusharora:
Issue #674
Consider I have not made it as it is like fossasia.org =>1.  Transparent on top 
                                                                                       2. Semi transparent on scroll.

Instead of this, Here it is only semi-transparent 
Reason:  We only have one logo image provided by user. Fossasia.org uses two logo images one light and other dark to handle it beautifully. Making topbar same as fossasia.org will require a light logo to be visible on overlay.
Image1 

Image2


championswimmer:
cool !

PR #682, opened 2016-08-22
Fix timezone offset

championswimmer:


championswimmer:
merging myself because the only change it does is change time offset by 2

PR #687, opened 2016-08-23
documentation of components and how final html is formed 

championswimmer:
[skip ci]


championswimmer:
merging myself as it has no change in code (only docs)

PR #713, opened 2016-08-27
Fix erroneous Line Breaks in Track Info Areas

### LINKED ISSUE #700: Erroneous Line Breaks in Track Info Areas
There are breaks right after the colored dots in the track info areas, that should not happen.

screenshot from 2016-08-26 15 12 03

shubham-padia:
Tries to fix #700 . 

The bug mentioned in the issue #700 was occurring because no explicit width was mentioned for the inline block. A new css class `titletext` is added to fix the issue. 140px was the maximum size i could go with, from 141px onwards the line breaks occur again.

***Before :*** 
selection_002

***After :***
selection_001

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -636,6 +636,12 @@ a {
     height: 12px;
     border-radius: 50%;
   }
+
+.titletext {
+  vertical-align: top;
+  width: 140px;
+}
+
   .title-inline { 
     padding:0;
     li{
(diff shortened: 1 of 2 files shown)


aayusharora:
@shubham-padia Thank you, but I have already solved this. I will merge this , if it looks good. Please comment on assigned issues before generating PR.

shubham-padia:
@aayusharora Okay :smile:

aayusharora:
Ok @shubham-padia  This is giving not the desired output. 
It is giving this

shubham-padia:
No idea why that occurs :smile: , looks like this on my pc
screenshot from 2016-08-28 03-46-38

aayusharora:
Ok, I have already solved this https://github.com/fossasia/open-event-webapp/pull/714 hence closing this.

Issue #723, opened 2016-08-28
Ensure left side bar in schedule and rooms page is same style


mariobehling:
As below the left area of the schedule page should use the same style/width like the rooms page.

Please change the schedule page to the same as the rooms page.

![screenshot from 2016-08-28 20 25 55](https://cloud.githubusercontent.com/assets/1583873/18035874/a3fa0cd8-6d5e-11e6-9be9-de0b661ca89e.png)

mariobehling:
@aayusharora Does not seem solved in latest test version. Left side bar still has different width.

championswimmer:
this is done now, and up on gh-pages. can be checked. 

http://fossasia.github.io/open-event-webapp/rooms.html
http://fossasia.github.io/open-event-webapp/schedule.html

PR #725, opened 2016-08-28
padding adjustment roompage

aayusharora:
Issue #723
I have find out very less difference between styling of left panel of Room and Schedule page. 
That was only padding.
The width of left panel for Schedule page is greater because Room page also contain Track bullets on right which requires their own space.
Corrected Padding here


mariobehling:
Please ensure that left panel is same width on both pages. Your example screenshot shows wrong times.

championswimmer:
@mariobehling correct times are shown when hosted on heroku

Issue #701, opened 2016-08-26
Footer menu links not working on all pages


mariobehling:
* [x] About page not correctly linked from all pages
* [ ] Ticket link should link to "ticket area" on start page
* [x] Scroll to top should use scrolling feature

![screenshot from 2016-08-26 15 15 15](https://cloud.githubusercontent.com/assets/1583873/18006474/fa905bbc-6b9f-11e6-8dec-3ccb619f593b.png)

shubham-padia:
@championswimmer Can i work on this ?

shubham-padia:
@championswimmer Can I work on this ?

shubham-padia:
@championswimmer I've implemented a solution for this issue, should I send a Pull Request for review ?

shubham-padia:
@mariobehling @championswimmer Please review #733 .

championswimmer:
@shubham-padia no it means ticket should link to index.html#ticketbutton

shubham-padia:
@championswimmer  please review #735 for ticket link.

PR #733, opened 2016-08-29
Fix footer menu links issue

shubham-padia:
Intends to resolves 2 out of 3 issues in  #701.
- fixes " About page not correctly linked from all pages ".
- fixes "Scroll to top should use scrolling feature"


championswimmer:
thank you

Issue #738, opened 2016-08-29
Rooms page - event dates - links not using scroll effect


mariobehling:
Tracks and Schedule pages are using the scroll effect but the rooms page date links do not.

* [ ] Please implement the scroll effect for date links 
http://fossasia.github.io/open-event-webapp/rooms.html

championswimmer:
deployed  http://fossasia.github.io/open-event-webapp/rooms.html

@mariobehling  you have to wait till the whole page loads before the scrolling functionality works (it is in js, so scroll.js must load, untill it loads, normal href-id linking works).

PR #739, opened 2016-08-29
Fix date link scroll

championswimmer:
fixed https://github.com/fossasia/open-event-webapp/issues/738


championswimmer:
merging myself, to be able to deploy. have tested with ots and fossasia both to work

PR #743, opened 2016-08-30
change color of navbar icon-bar and add some space before it

alexagr:
cosmetic fixes to navbar in mobile view:
- change color of icon-bar - so that it's more visible on semi-transparent background on HOME page
- add a bit more space between social icons and icon-bar - so that it doesn't look too "crowded"


championswimmer:
hey thanks  !

PR #744, opened 2016-08-30
fix Tracks screen to work with unicode track names

alexagr:
fix Tracks screen so that it's correctly displayed when track names are Unicode
the problem is that slugify() removes all non-ASCII characters - essentially making slug contain date only; as slug is used to only internally group sessions, removing slugify() call fixes the problem


alexagr:
c7a7398 - fix dates order on Tracks and Rooms screens
In FOSSASIA example you have 3 dates - Friday, Saturday and Sunday - that accidentally are alphabetically sorted. However if you change dates, for example, to Thursday, Friday, Saturday - you will get them displayed in wrong order on Tracks and Rooms screen. This is because corresponding maps are sorted by **date** field which starts with "dddd". Instead I've created sortKey field - and sort by it instead. Problem solved.

aayusharora:
Thanks @alexagr  Looks good :)

PR #745, opened 2016-08-30
fix timezone handling for session start/end

alexagr:
Session start/end time is encoded in JSON files in UTC format.
Therefore correct way to transform it into local time is as follows:
    moment.utc(<time from JSON>).localtime().format("HH:mm")

Before my fix, there was some "crazy" code that used hardcoded offset (of 4) that applied "on top" of localtime offset (applied by mistake). Moreover offset was applied to time calculations only - but not to dates... Anyway - now it should be clean and clear.

### DIFF
--- a/src/backend/fold.js
+++ b/src/backend/fold.js
@@ -48,7 +48,7 @@ function foldByTrack(sessions, speakers, trackInfo, reqOpts) {
     }
 
     // generate slug/key for session
-    const date = moment(session.start_time).format('YYYY-MM-DD');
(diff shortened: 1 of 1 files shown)


mariobehling:
Thanks for the pr. Could you resolve the merge conflict, please? Thank you.

alexagr:
sorry - I'm pretty new to the GitHub - how exactly can I resolve the merge conflict?

alexagr:
Merge conflicts resolved.

mariobehling:
Ah, ok that was fast :-)

PR #748, opened 2016-08-30
beautify speakers list without biographies

alexagr:
For speakers that lack biographies:
- change `"About <speaker name>:"` to `"<speaker name>"`
- make horizontal separator below the speaker visible


mariobehling:
Thank you!

PR #756, opened 2016-08-31
fix sessions grouping in Rooms page

### LINKED ISSUE #741: Rooms page incorrectly groups sessions for the 2nd and 3rd day
Rooms page works just fine for the 1st day.
But for the 2nd and 3rd day sessions are grouped incorrectly - e.g. "Google Summer of Code" session that happens in Dalton Hall is shown under Lewis Lab.

alexagr:
resolves #741 

There was a bug in foldByRooms() implementation that relied on certain order of sessions in JSON file to properly group them. As a result of that some sessions on 2nd and 3rd day of FOSSASIA example were "assigned" to wrong rooms. This commit fixes the issue.

### DIFF
--- a/src/backend/fold.js
+++ b/src/backend/fold.js
@@ -522,18 +522,16 @@ function foldByRooms(room, sessions, speakers, trackInfo) {
     if (room == undefined) {
       return;
     }
+    
+    let venue = '';
     if(session.microlocation !== null ) {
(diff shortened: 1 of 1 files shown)


mariobehling:
This is really helpful and we should have seen that much earlier. Thank you!

Issue #753, opened 2016-08-30
Header problems on mobile device


alexagr:
- when navbar is wider than a single line (as in FOSSASIA example that has many social links) it overlays the page title
- when days do not fit on a single line (as in FOSSASIA example on FHD screen) the day displayed on the 2nd line partially overlays the day displayed on the 1st line

aayusharora:
Thanks @alexagr It is solved.

PR #758, opened 2016-08-31
Mobile layout

### LINKED ISSUE #752: Rooms page is too wide on mobile device
When viewed on mobile device (Android 5.1, latest Chrome version, FHD resolution) Rooms page is too wide - so there is horizontal scroll and icon-bar is not visible unless you scroll to the right. This happens only on Rooms page - other pages are rendered just fine (without horizontal scroll).

alexagr:
layout fixes on mobile devices
resolve #752 and #753

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -678,7 +678,7 @@ a {
 }
 
 .track-names {
-  color: #000000;
+  color: black;
   margin-top: 16%;
 }
 
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -415,11 +415,16 @@ a {
 
 }
 
+.date-list {
+  line-height: 40px;
+}
+
 .date-pill {
   border: thin solid;
   border-radius: 4px;
   margin: 10px;
   padding: 6px;
+  white-space: nowrap;
 }
 
 #track-list { 
@@ -607,6 +612,11 @@ a {
   width:299px;
 }
 */
+
+.track-heading {
+  margin-top: 40px;
+}
+
 .active {
   color: $black !important;
   font-weight: bold;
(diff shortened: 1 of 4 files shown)


alexagr:
Sorry for too many commits - I fixed alignment of some existing block and got hound for errors that were in it before I event touched it; hopefully they are fixed now.

src/backend/_scss/application.scss
```diff
@@ -664,20 +673,25 @@ a {
...
- @media (max-width: 990px) {
-    .title-inline {
-      width: 50%;
-      float: left;
-    }
-    .track-names {
-      margin-top :0;
-    }
+.track-names {
+  color: #000000;
```
aayusharora: Please use config.scss to use variable color name instead of hex color value

alexagr: sorry - misinterpreted the hound report - fixed in the latest commit

alexagr: right - I misinterpreted the hound report - fixed in the latest commit

alexagr:
Hugh - for some reason my line comment disappears - already wrote it twice... Anyway - the problem with "black" color is fixed.

src/backend/_scss/application.scss
```diff
@@ -664,21 +673,26 @@ a {
...
- @media (max-width: 990px) {
-    .title-inline {
-      width: 50%;
-      float: left;
-    }
-    .track-names {
-      margin-top :0;
-    }
+.track-names {
+  color: black;
```
aayusharora: @alexagr it should not be 'black', it should be $black that is a sass variable in config.scss.

aayusharora:
@alexagr Please see the above comment.

alexagr:
oh - that makes sense - fixed

aayusharora:
Looking Nice @alexagr Merging this.

PR #761, opened 2016-08-31
Fix alignment issues in schedule page popover

### LINKED ISSUE #760: Schedule Page: Align name, organization left and track left
In some long name cases name and organization could look broken. Also alignment of track should be left.

On schedule page please align:
* [ ] name left
* [ ] organization left
* [ ] track left

screenshot from 2016-08-31 22 59 05

shubham-padia:
Intends to resolve #760 .

result:

after


mariobehling:
Nice work. Thank you!

PR #782, opened 2016-09-06
updated site

mariobehling:
Last version of fossasia site.


json/speakers
```diff
@@ -4731,19 +4695,20 @@
...
         "photo": "/images/speakers/JonPhillips_175.jpg", 
         "organisation": "#NEWPALMYRA", 
         "position": null, 
         "country": "USA", 
         "short_biography": "", 
-        "long_biography": "Barry Threw (@barrythrew) is an artist and technologist focused on spatial media, mutible architectures, and cultural infrastructures. He works in collaboration with institutions, artists and organizations at the intersection of technology and culture. He is Director of Software at Obscura Digital, and is a curator with the Gray Area Foundation for the Arts. He works on free software and open knowledge projects with Fabricatorz.  He is serving as Interim Director of #NEWPALMYRA until Bassel Khartabil's release.", 
+        "long_biography": "Jon Phillips.", 
         "website": "http://newpalmyra.org", 
-        "twitter": "https://twitter.com/NewPalmyraOrg", 
+	"website": "http://rejon.org",
```
alexagr: - [ ] delete previous "website" attribute
- [ ] fix alignment

json/speakers
```diff
@@ -3482,16 +3445,17 @@
...
+        "photo": "/images/speakers/AnhTuanTruong_158.jpg", 
         "organisation": "VFOSSA", 
         "position": null, 
         "country": "Not available", 
         "short_biography": "", 
         "long_biography": "Start working after being graduated since 1995, I have worked as almost roles in software industry, from software developer, team leader, project manager, product manager to general managers like CTO, CIO and Company president.  I love the freedom in free and/or open source software so I spend almost my time on working together with friends and colleagues to develop, promote FOSS and FOSS communities more and more.  I am currently member of many FOSS communities, domestic as well as international, generic as well as product specific, such as: HanoiLUG, AsiaSource 2, Zimbra, OTRS, Drupal, Fedora, CentOS linux, etc. I am currently Vietnam Free and Open Source Software Association Vice President.", 
-        "website": "http://vfossa.vn/", 
-        "twitter": "", 
+        "website": "http://vfossa.vn", 
+        "website": "http://www.iwayvietnam.com",
```
alexagr: "website" attribute appears twice

mariobehling:
@alexagr We intentionally did not make a specification, that there needs to be one attribute only. People can have several websites, blogs, twitter accounts etc. 

Why should this be unique?

mariobehling:
@alexagr While this is handled in this way in the orga server and early version of FOSSASIA/OpenTechSummit event sites, I see that there is only one link generated. I filed an issue here https://github.com/fossasia/open-event-webapp/issues/784 therefore.

alexagr:
@mariobehling 'cause you wanna be able to parse it using standard JSON library, and most libraries don't support multiple keys

mariobehling:
@alexagr Yes, good point. However, the multiple link case is a usual case and in the orga server we left the naming to the user (in the same way that users can add custom items in Gmail contacts).

I assume a standard library would simply use the first or last entry then and this would mean we lose the others. And if you stick to one entry you will not have a problem anyways.

Issue #787, opened 2016-09-07
Json parsing converts html tags to escape sequence


Princu7:
Screenshot of problem caused due to it.
![screenshot from 2016-09-08 02-12-48](https://cloud.githubusercontent.com/assets/8847265/18327857/0a53f06c-756a-11e6-9547-f1961049c140.png)

PR #788, opened 2016-09-07
Update rooms.hbs

Princu7:
This is in reference to the issue #787 .  Used triple mustaches to unescape html.
screenshot from 2016-09-08 02-28-43

### DIFF
--- a/src/backend/templates/rooms.hbs
+++ b/src/backend/templates/rooms.hbs
@@ -112,7 +112,7 @@
                       {{/if}}
                       {{#if long_biography}}
                       <p class="session-speakers-more">
-                        {{long_biography}}
+                        {{{long_biography}}}
                       </p>
                       {{/if}}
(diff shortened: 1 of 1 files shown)


shubham-padia:
@Princu7 Not escaping user input may lead to xss vulnerabilities. Although something like this : https://github.com/punkave/sanitize-html  or https://github.com/chriso/validator.js can be used to sanitize user input to allow certain html tags.

Princu7:
Ah. I see. Thanks for explaining. So the issue #787 isn't really a bug but a fault in the input json file which automatically assumed that the html will not be escaped. So should I close the issue and pull in the request back?

aayusharora:
@Princu7 Thanks :) Looks Good.

shubham-padia:
@aayusharora Won't this lead to XSS vulnerabilities?

Issue #789, opened 2016-09-08
Links not working on Tracks page Speakers name to redirect to particular Speaker


aayusharora:
http://fossasia.github.io/open-event-webapp/speakers.html#1 Links like that are not working now.
Steps To Reproduce 
1. Go to sample site tracks page http://fossasia.github.io/open-event-webapp/tracks.html
2. Click the name of any speaker like Mario Behling (OpenTech Society) in the card. The link http://fossasia.github.io/open-event-webapp/speakers.html#2 will redirect on Speaker page but not point to exact Speaker.

mariobehling:
Where is that needed? Please try the latest version and feel free to upload the latest generated version. No need to submit issues for old versions.

aayusharora:
@mariobehling Added Steps to reproduce

tadityar:
I can confirm this issue, the link in the speaker's name does not point to the correct anchor (the picture does, though). Will create a PR to resolve this.

aayusharora:
@tadityar Thank you Please go ahead :)

aayusharora:
Done !!

PR #793, opened 2016-09-09
Fix tracks page speaker link

tadityar:
This PR fixes issue #789 by changing the link to use nameIdSlug instead of id.

Oh, and my text editor removes trailing whitespaces. Should I just leave it there or remove it (like the changes in this PR)


aayusharora:
Thank you @tadityar

PR #812, opened 2016-09-15
closed tags and removed redundant closing tags

jainaman224:
I have removed redundant closing tags and added closing tags for unclosed html tags. Review this and please suggest improvements.


jainaman224:
screenshot from 2016-09-15 17-58-34

Now we have underline against current page not just under home.

Issue #815, opened 2016-09-15
Bad Layout Speakers page on Mobiles


aayusharora:
By some previous PR's , Speaker page layout is disturbed on Small screens.
![4](https://cloud.githubusercontent.com/assets/12194719/18566800/f984df04-7bb2-11e6-84f2-15470c036fea.png)
- [ ] Shift popover exactly below the image.
- [ ] Remove X-axis scrollbar

Princu7:
In the current implementation, on small screen the pop-up box is displayed from the centre of the image. So we have added a class which adds a necessary left margin to make the pop-up box shift to the left to make it coincide with the image.  The problem with that is that the different small screen sizes will require different left margin for perfect overlap. So we have to add various more screen sizes and the corresponding left margin for making the pop up come directly under the screen. The previous pull request left out some cases so it was working on devices with width size 320-360px but looked a bit abrupt on 500px wide screens.There might be better solutions but we will have to think for those. I am willing to work on this. Just need some guidance.

Princu7:
@aayusharora Not exactly the same thing as asked, but please do check out this page I made after doing some changes . Run it on different screen sizes and it should not look bad in general. 
https://princu7.github.io/websiteTest/speakers.html
There are some cases where we don't want to shift the popovers exactly below the image. Suppose the width of the screen is 700px then only one picture of width 300px will come in a single row. If we shift popovers exactly below the image, then a lot of the white space will be wasted(400px to be precise). A better alternative than that would be just to display the picture in the middle and let the pop-over have the full size of the screen. It does look neat and clean.
For example-
Intended version
![screenshot from 2016-09-16 15-52-17](https://cloud.githubusercontent.com/assets/8847265/18583028/dedf04cc-7c25-11e6-808a-bb35c1d304d5.png)

Alternative version
![screenshot from 2016-09-16 15-53-07](https://cloud.githubusercontent.com/assets/8847265/18583043/eb180040-7c25-11e6-9a4a-de066b200a6a.png)

The second looks a tiny bit better. Views and suggestions?

aayusharora:
@Princu7 The second view is destroyed badly. To improve 
Follow the first view.

Princu7:
@aayusharora Ah. My bad. No problem. I will implement keeping the first view in mind. Will Ping you as soon as I am done with it. Thanks for the feedback.

Princu7:
[...]

PR #833, opened 2016-10-04
Made some changes to the application.scss file

Princu7:
Solves the issue #815 .
@mariobehling  @aayusharora 
Here is the link to the gh-pages
https://princu7.github.io/websiteTest/speakers.html

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -484,7 +484,7 @@ a {
 .arrow {
   height: 10px;
   left: 6px;
-  margin-top: -2.3%;
+  margin-top: -9px;
   overflow: hidden;
   position: absolute;
   width: 25px;
@@ -803,12 +803,13 @@ width: 116px;
 
 */
 
-@media (max-width: 544px) {
+@media (max-width: 767px) {
   .pop-box {
-    width: 350px;
+    width: 100%;
 
     .arrow {
-      margin-top: -2.5%;
+      left: 48%;
+      margin-top: -10px;
     }
   }
 }
@@ -1043,19 +1044,11 @@ width: 116px;
   span.speaker-name {
     text-align: center;
   }
-
-  .pop-box {
-    left: 100px;
-  }
 }
 
 @media only screen and (max-width: 500px) {
   span.speaker-name {
     text-align: center;
-  }
-
-  .pop-box {
-    left: 0;
   }
 }
 


mariobehling:
The result of this change is, that in normal view the width of the dropdown speaker info is reduced. There is no reason to reduce the width in larger resolutions.

screenshot from 2016-10-04 14 17 46

Larger width as right now is fine (see below).

screenshot from 2016-10-04 14 19 26

mariobehling:
A problem I see in this dropdown box is also that the line around the box is missing at the corners.

Princu7:
@mariobehling Thanks for pointing out the issues. I will correct them right away.

Princu7:
I made some changes. Please have a look at my gh-pages and review it. Thanks.
https://princu7.github.io/websiteTest/speakers.html

mariobehling:
Thank you. On the top right and bottom right, the corner lines are still not completely closed. Is is possible to fix this?

Princu7:
@mariobehling I am sorry. I tried hard to look for the cause of it but couldn't find it.  I have updated my pull request to the previous point( having the changes you mentioned). Suggest further course of action.

src/backend/_scss/application.scss
```diff
@@ -481,7 +481,7 @@ a {
 .arrow {
   height: 10px;
   left: 6px;
-  margin-top: -2.3%;
+  margin-top: -9px;
```
aayusharora: Why it is not relative? Why to change this to px. Any reason?

Princu7: @aayusharora I am sorry. While working with other issues, the changes made in the gh-pages due to this pull request got lost and what you saw was the previous version without any alteration. I have updated my gh-pages to contain the changes done in this pull request.
The X-axis scroll bar is almost gone. It's still there but it's range is almost negligible.  Meaning that you can easily view the whole content without having to scroll to the right. To completely remove it we have to add the following properties to the code
```
body {
 overflow: -moz-scrollbars-vertical;
 overflow-x: hidden;
 overflow-y: scroll;
} 
```
I wasn't too sure about adding it so left it up for your review. About the margin values for the arrow, I tried different values both relative in percentages and absolute in pixels and the -10px seemed to work well on all the screens. We can change it if there is some issue with it. Thanks.
Please review again.
Link to updated gh-pages
https://princu7.github.io/websiteTest/speakers.html

mariobehling:
@aayusharora Can this PR be merged?

mariobehling:
Merging. Let's follow up on open questions in future PRs.

PR #842, opened 2016-10-09
Fixing border corners issue in speaker popups

poush:
From #833 by @mariobehling 
`Thank you. On the top right and bottom right, the corner lines are still not completely closed. Is is possible to fix this?`

This pull request fixes that corners issue and small whitespace inside border corners.

screen shot 2016-10-09 at 6 10 58 pm


mariobehling:
Thank you.

Issue #845, opened 2016-10-13
Navbar links don't highlight properly after clicking on session of a speakers or vice versa


Princu7:
@mariobehling @aayusharora 
We are on the speaker page and click on one of his/her sessions. Since we are on the speakers page currently, the _speaker_ link in the navbar  is underlined.
After clicking on the session link we are redirected to the tracks page. But instead of the '_tracks_', '_home_' is highlighted in the navbar.
Some screenshots demonstrating the issue
Before the click
![screenshot from 2016-10-13 09-41-33](https://cloud.githubusercontent.com/assets/8847265/19336557/0df040b4-912a-11e6-8de7-ea20553f2e52.png)
After the click
![screenshot from 2016-10-13 09-41-55](https://cloud.githubusercontent.com/assets/8847265/19336566/1a78e656-912a-11e6-9fc5-6c179fb3c297.png)
I will solve this issue along with #830.

Issue #830, opened 2016-09-26
Problem with speaker  and the session links


Princu7:
On clicking on the speaker pictures on the schedule, rooms and the tracks page, it just takes us to the speakers page. That is, it doesn't matter which speaker picture you want to click it will take you to the top of the speaker page rather than the desired behaviour of taking us to the actual position where the speaker that we have clicked is situated.
Also on the speakers page, when we click on the sessions to be delivered by him/her, just like in the previous case, it  takes us to the top of the tracks page instead of taking us to the proper place where that session is mentioned in the tracks page.
Screenshots demonstrating the issue
No matter where we click
![screenshot from 2016-09-26 21-06-59](https://cloud.githubusercontent.com/assets/8847265/18840753/45593934-842d-11e6-9a50-3c8e0feab8db.png)
We get taken to the top of the speakers page
![screenshot from 2016-09-26 21-05-32](https://cloud.githubusercontent.com/assets/8847265/18840841/894e95c6-842d-11e6-9b72-13cbc278737d.png)
@mariobehling  @aayusharora 
I have started to work on the issue. Will send a request soon.

Princu7:
@mariobehling  @aayusharora  Sorry for the delay in dealing with this issue. The problem here is that we are using the name tag in the anchor element of the speaker and the sessions in order to link  them. But the <a> element doesn't support name attribute in HTML5 and we will have to use id instead of them. That's why it is not linking properly. I will implement it and send in a request.

PR #847, opened 2016-10-13
Fixes the problem with the session and the speakter links. Also fixes navbar links.

Princu7:
Solves the issue #845 and #830.
https://princu7.github.io/websiteTest
@mariobehling  @aayusharora 
Please review it. Thanks


mariobehling:
Good job! Thank you!

Issue #751, opened 2016-08-30
Implement a more standard conform generated site


mariobehling:
Currently the site has a folder "dependencies" with different files. When the site is generated, please move files to more standard conform folders as follows.

From dependencies folder in generated sites move:
* [ ] css to `/css`
* [ ] the folder `fonts` to into the root directory
* [ ] js files into `/js`
* [ ] avatar.png into the folder `images`

Finally, please
* [ ] delete folder/do not generate folder "dependencies"

Princu7:
@mariobehling  Can I take this up?

Princu7:
@mariobehling Please test the app here
https://tranquil-temple-31388.herokuapp.com/

shubham-padia:
@mariobehling Looks like this issue has been resolved via #851. Please update the issue. Thank you.

PR #851, opened 2016-10-15
Implemented more standard conform generated site

Princu7:
@mariobehling  @aayusharora  Solve the issue #751
Test the app here
https://tranquil-temple-31388.herokuapp.com/


mariobehling:
Thank you. Please review hound-ci errors.

Please provide a link to a generated sample site.

Princu7:
@mariobehling  I have reduced the no of hound violations. They are arising due to the use of _console.log()_ statements in the code. I have removed all the not so important _console.log()_ statements and rest are kept and are in accordance with the previous structure of the code.

In this issue, the objective was only to remove the _dependencies_ folder in the generated event folder  and to move its contents to other folders. There were no front-end improvements so website still looks the same as before it looked in this branch before this commit. You can test that the dependency folder is not being generated by generating the event folder and downloading it through the link given below-
https://tranquil-temple-31388.herokuapp.com/

https://princu7.github.io/websiteTest/
This is the sample website. It looks the same as before. It might be slightly outdated since I haven't merged the latest commits in this branch. But this request doesn't cause any merge conflict so there is nothing to worry about.

Please test the app and review. Thanks.

mariobehling:
Thank you. Good job.

Issue #718, opened 2016-08-28
On tracks page implement same style as on rooms page


mariobehling:
We had started with the style on the tracks page, but it has evolved since then. Please implement the same style on the tracks page therefore as on the rooms page to keep the app style consistent.

(First solve the design issues/bugs on the rooms page, please.)

jainaman224:
@mariobehling Can I work on this issue please...?

jainaman224:
@mariobehling Sir, where should I add video, audio and slide section on tracks page and I am performing clean up for application.scss .

mariobehling:
@jainaman224 Please propose where it could be suitable.

aayusharora:
@mariobehling I think the issue needs to be updated .

PR #853, opened 2016-10-16
Issue718

jainaman224:
@mariobehling Please review at https://jainaman224.github.io/open-event-webapp/issue718/tracks.html regarding #718 .
Thank you.

### DIFF
--- a/src/backend/templates/tracks.hbs
+++ b/src/backend/templates/tracks.hbs
@@ -4,17 +4,17 @@
   from https://github.com/fossasia/open-event-scraper.
 -->
 <html lang="en">
-<head>
+  <head>
 
-  <meta charset="utf-8">
-  <meta http-equiv="X-UA-Compatible" content="IE=edge">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


mariobehling:
Thank you. Looks good, but why do the times not match anymore, e.g. Grand Opening starts at 15.30.

jainaman224:
@mariobehling Sir, I think time is correct because time for Grand Opening is 9:30 UTC and time zone for my country is UTC+5:30.

mariobehling:
@jainaman224 Yes, I understand your idea, but the timezone of a real-world event should be the local timezone where the event takes place. Please also resolve merge conflicts. Thank you!

jainaman224:
@mariobehling It will be better to get local time rather than UTC time in json itself. It will be faster.

mariobehling:
Thanks for your ideas, but we had this discussion before and for now the current implementation works best across projects. Cheers

mariobehling:
@jainaman224 Let's progress. I am merging this and we continue from there. Thank you.

jainaman224:
@mariobehling Thank you.

Issue #836, opened 2016-10-05
Generator-Page: Take out "App's Name" and generate name from event JSON


mariobehling:
The Android app generator has implemented this already.

The goal is to reduce the form by one field. Instead of having the user entering a name for the app. The app's name should be taken from the event JSON, e.g. https://github.com/fossasia/open-event/blob/master/sample/FOSSASIA16/event

Princu7:
I would like to work on this one!!.

Princu7:
@mariobehling  @aayusharora  I have done some work on this issue. Some things which I want to mention-
Current workflow-

1. We are entering the email address and name of the app in the form. Then we are checking in the dist folder for the same name and if found, deleting it.

2. Making the folder of the same name as entered in the form, and copying the uploaded files there. Then we extract information from the uploaded json files and compile the sass  and the template files.

Now since we are not entering the name of the app in the form itself, we have to first give a temporary name for the folder something like , _"tempProjectName"_ , since we will know the actual App's name only after we have copied the json files into the folder and then extracted name from event json file. Now after extracting the actual app's name from the event json file, we will check for the folder of the same name in the working directory and clean it unlike the previous case where we clean the folder in the starting itself. 

Proposed workflow-
1.  User only enters the email address

2.  Let the temporary name of the project be "tempProject", create a folder of the same name and  extract the uploaded files there. Directory structure looks something like
user-email@address.com
    tempProject

3.  Extract information from the json files and save the actual name of the project in a variable.

4.  We check for the existing folders of the same name as that of the actual project name and if found, delete it.
So if the actual name of the project was 'FOSSASIA 16' and we already had a folder named the same in the working directory, then delete it.

5.   Rename the 'tempProject' to the actual project name.

mariobehling:
@Princu7 Please have a look at the Android app generator, if we can align the process here. It is already implemented in a way here, where the App name field is not required: https://github.com/fossasia/open-event-android

> Rename the 'tempProject' to the actual project name.

Why is this necessary?
The temp project folder solution sounds good, but the directory structure is not so important as we don't use the directory listing anywhere. Or am I missing something? This is just for the upload of the zip, correct? The event folder will only be generated in the following steps.

Princu7:
@mariobehling It seems like some misunderstanding have happened. Apologies for that. The whole process described above is not for the upload of the zip but for the event folder itself. Let me try to explain-

1. The uploaded zip is copied in a folder named '_upload_' in the project directory.

[...]

PR #862, opened 2016-10-18
Removed the option to enter the app's name in the form and instead us…

Princu7:
…ed the name given in the event.json file to generate the event folder. Also fixed some code indenting issues.
Solves the issue #836. 
@mariobehling 
Test the app here
http://fierce-brushlands-59495.herokuapp.com/
Please review. Thanks.


mariobehling:
Good job. Thank you!

Issue #866, opened 2016-10-18
UI Bugs - sidebar line and scroll feature on clicking dates


mariobehling:
* [x] Ensure left sidebar line is going through without breaks on rooms and tracks page

![screenshot from 2016-10-18 23 45 11](https://cloud.githubusercontent.com/assets/1583873/19497553/21010ed8-958d-11e6-9cc3-7811e8c31d7a.png)

* [x] On tracks and rooms page ensure dates clicked use the scroll feature, e.g. test Friday and Sunday on sample site http://fossasia.github.io/open-event-webapp/tracks.html

Princu7:
I can work on this one.
Regarding first point, does this look good? @mariobehling 
![screenshot from 2016-10-19 14-17-47](https://cloud.githubusercontent.com/assets/8847265/19511929/30ee4d68-9607-11e6-8b92-5a73bf4fa036.png)

Princu7:
@mariobehling  I have completed the work. The second bug was quite nasty and took quite some time to find. The problem was we were using the _slug_ of a track as an _ID_ but the slug can sometimes consist of multiple words consisting of spaces which isn't the usual norm for naming an _ID_.Anyways I made some changes in the scroll.js file and it is working fine now. You can test it here.

https://princu7.github.io/websiteTest/
Please review. Thanks!!

shubham-padia:
@mariobehling Looks like this issue can be closed. Please update the issue. Thank You.

PR #868, opened 2016-10-19
Solved the line issue and the scroll issue

Princu7:
Fixes the issue #866 . 
Test it here
https://princu7.github.io/websiteTest/
Please review. Thanks!!

### DIFF
--- a/src/backend/assets/js/scroll.js
+++ b/src/backend/assets/js/scroll.js
@@ -1,14 +1,17 @@
       $(function() {
       $('a[href*="#"]:not([href="#"])').click(function() {
         if (location.pathname.replace(/^\//,'') == this.pathname.replace(/^\//,'') && location.hostname == this.hostname) {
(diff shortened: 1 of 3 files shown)


mariobehling:
Thank you. Please check about hound-ci issues. Cheers.

Princu7:
@mariobehling
 I have made some changes but the houndci issues are nothing to worry about. They warn about the not defined $ symbol. Since the $ symbol comes from the jquery library which will also be included in the file in which we are including this very file itself so there will be no problems. Also the rest of the code is in the same veins with the previous structure of the code.

mariobehling:
Yes, I know, but we should tackle some of these as well. Let's proceed step by step. Thank you.

Princu7:
@mariobehling Makes perfect sense. Focus on big issues and then shoot down all the smaller ones. :+1:

PR #869, opened 2016-10-20
Auto deploy sample site to heroku

### LINKED ISSUE #861: Auto-deploy sample of Open Tech Summit to Heroku
The sample of the OpenTechSummit has a lot of audio files, that are too big to deploy to Github without Large File Support. Therefore please auto-deploy the sample on pull request merges to Heroku.

The sample file is here: https://github.com/fossasia/open-event/tree/master/sample/OTS16

shubham-padia:
resolves #861 .

@mariobehling It looks like the `OTS16` sample is broken. Currently i have implented auto deploy to heroku using the `FOSSASIA16` sample itself, you can change it afterwards by changing the `HSAMPLE_ENDPOINT` and `HSAMPLE_EVENT
you can see the demo here : http://shubhamopevwebgen.herokuapp.com/sample/

Following env variables are provided for further customizations , you can set them in the travis settings if you want to change something from its default value.

- `HSAMPLE_ENDPOINT` : default is the FOSSASIA16 sample. 
- `HSAMPLE_TIMEOUT` :default is 60s same as `GH_TIMEOUT` but for sample to be deployed to heroku.

As of now, app names are set from the json itself instead of the previous approach.
So , two more variables have been added to cope up with that.
- `GH_EVENT` - event name specified  in the json along with proper escaping according to unix .
- `HSAMPLE_EVENT` - event name specified  in the json along with proper escaping according to unix . 
- Default to both of the above variables have been set to "FOSSASIA\ 16"


mariobehling:
@shubham-padia Thank you, but we already have autodeployment for FOSSASIA to the github branch. Please help us to make the OTS sample work and open issues if anything is not working. Cheers.

Issue #819, opened 2016-09-21
Add error information in case build fails


mariobehling:
![screenshot from 2016-09-21 15 55 27](https://cloud.githubusercontent.com/assets/1583873/18713985/92123b58-8014-11e6-98a7-b374cc52efb2.png)

Please add a simple triangle in the area "copying assets". 

Expected behaviour: After the login is finished or failed the user can see the log info.

There can be a triangle button with the log info that can be expanded by the user, if wanted, something like "> See build logs"

Princu7:
Can I take this up @mariobehling

mariobehling:
@Princu7 Yes, sure. Please post screenshot on how you implement it here. Thank you.

Princu7:
@mariobehling I have done some work on this issue. First I have classified the log messages into three types  - 
'Info' - Describes some task which is being done or started.
'Success' - Describes the successfull completion of an event or task.
'Error' -  Describes a task that has failed. It will have the option of viewing the detailed information about the error.

The logs are generated on the go i.e. user can view the log while the app is being generated in real time.I will add some screenshots and the working demo app as soon as I am finishing it. Please give some suggestions and advice. Thank you

PR #879, opened 2016-10-26
Added logging support for the app

Princu7:
Solves the issue #819. @mariobehling @aayusharora  Please test and review the app at
https://floating-tor-16309.herokuapp.com/

Here are some screenshots
screenshot from 2016-10-27 01-42-26
screenshot from 2016-10-27 01-42-17
An error which I created myself. You have the option to view the detailed information of an error
screenshot from 2016-10-27 01-50-36
screenshot from 2016-10-27 01-50-41

The logging is not perfect but it is a good start for us. We will keep improving it as the time progresses. Please give you suggestions and reviews. Thanks.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -72,7 +72,7 @@ app.use('/', express.static(__dirname + '/www'));
 app.use('/live/preview', express.static(__dirname + '/../dist'));
 
 app.get('/download/:email/:appname', function (req, res) {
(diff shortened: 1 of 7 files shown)


mariobehling:
Thank you.

* [ ] Please make the build logs into a simple link - Logs - (not a button), that expands if click (just like now).
* [ ] Also right now, you show the logs right away: Only show the logs when people click on "Logs"

Issue #884, opened 2016-10-30
Default avatar not displayed on tracks pages


shubham-padia:
![selection_009](https://cloud.githubusercontent.com/assets/13910561/19833732/972ab2fe-9e6a-11e6-94f0-6e38bac069d3.png)

aayusharora:
Thanks ! I will look at it.

shubham-padia:
Nevermind :smile: , I've fixed it in #885 .

shubham-padia:
Closing this via #885 .

PR #885, opened 2016-10-30
Implement thumbnails and display default avatar for tracks page

### LINKED ISSUE #770: Implement two image versions for speakers and sponsors: Icon and Thumbnail
The orga server already offers two image sizes a) Icon and b) Thumbnail.

We also need two versions in the web app site.

We also need two versions in the web app site for the different areas.

In order to achieve this we could implement/add the different files sizes into the sample file. Alternatively we can scale the image sizes in the generator, which might be the more suitable for the time being in order not to clutter the file format too much.

shubham-padia:
resolves #770 and #884.
Head on to  https://shubham-padia.github.io/open-event-webapp/ for a preview.

@mariobehling please review.


mariobehling:
Thank you. Very good.

Issue #864, opened 2016-10-18
UI Improvements on small screens and mobiles


mariobehling:
- [x] On small screens ensure the search bar is using the full width
  ![screenshot from 2016-10-18 23 52 22](https://cloud.githubusercontent.com/assets/1583873/19497729/f30635d4-958d-11e6-9539-96284c043f7c.png)
- [x] On small screens ensure the date buttons are placed next to each other
  ![screenshot from 2016-10-18 23 55 23](https://cloud.githubusercontent.com/assets/1583873/19497820/69746740-958e-11e6-9743-39030edb625a.png)
- [x] On top menu bar show make menu item to use gray background on scroll over and click
  ![screenshot from 2016-10-19 00 03 07](https://cloud.githubusercontent.com/assets/1583873/19498047/6de8a614-958f-11e6-8dbe-ab4eb8d1bc06.png)

poush:
Search bar will be automatically fixed once #841 is merged. For other two things here is the screenshot of changes.

Grey Background on mobile screens
<img width="707" alt="screen shot 2016-10-29 at 11 47 23 pm" src="https://cloud.githubusercontent.com/assets/1812082/19831855/451d7f88-9e32-11e6-930a-7d3d2d0f8c9e.png">

and date buttons as
<img width="641" alt="screen shot 2016-10-29 at 11 47 08 pm" src="https://cloud.githubusercontent.com/assets/1812082/19831861/57e98954-9e32-11e6-9b43-58047d2dd591.png">

PR #886, opened 2016-10-30
Search Box and date UI fix 

aayusharora:
Issue  #864

d1


alexagr:
Sorry for disappearing for a while :)
IMO putting dates on 3 separate lines grabs too much screen real estate on mobile. We should instead put all 3 dates on the same line. This can be done, for example, by using shorter Day of Week name - e.g. "Mon" instead of "Monday".

aayusharora:
@alexagr Mon has changed to Monday ( full names due to some reason ) earlier, I myself think there is a space wastage in terms of UI/UX. But making them inline on small Viewports make them difficult while clicking i.e that's why col-md-12 layouts are preferred on mobile. 
Please consider all mentioned while concluding the best approach.
Thanks :)

alexagr:
@aayusharora In my opinion if we have to choose between short day names ("Sun" instead of "Sunday") on a single vs multiple lines - the former is a better trade off.

mariobehling:
Actually, it should be still ok, if the full days are written.

Would it also be an option to abbreviate the days only below a certain resolution?

aayusharora:
The day name generation format is handeled by the functions in `moment.js`. According to me, it's not feasible to that. @mariobehling Merge this if looks good.
I will come up with a solution to make it better on small screens.

Issue #890, opened 2016-11-01
Local upload of the OTS 16 Sample doesn't generate the event folder


Princu7:
@mariobehling  @aayusharora  When we try to upload the zip file of the OTS 16 sample, the event folder is not generated and the web app crashes. This is because of the large size of the OTS16 sample owing to which all the contents are not copied before they are read.
Here is an piece of the error 
`Error: ENOENT: no such file or directory, open '/home/princu/open-event-webapp/src/backend/../../dist/copy@gmail.com/tempproject/json/sessions'`

I am working on this issue.

aayusharora:
Thanks @Princu7

PR #892, opened 2016-11-01
Fixing the problem with the zip upload of the OTS16 sample

Princu7:
@mariobehling  Solves the issue #890 . App Used to get stuck or throw an error on upload of large sample zips like OTS16.  Here are the screenshots
screenshot from 2016-11-02 02-15-03
Now the local upload of the sample works!!
screenshot from 2016-11-02 02-02-16
screenshot from 2016-11-02 02-02-07
Since the whole sample site can't be run on github, here is a working demo of the site created from the zip upload of the OTS sample from the local machine
https://princu7.github.io/OTS/
Also refactor some old code and fixes indentation issues. Please review. Thanks!!

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
   },
   "homepage": "https://github.com/fossasia/open-event-webapp#readme",
   "dependencies": {
-    "adm-zip": "^0.4.7",
     "archiver": "^1.0.0",
     "async": "^2.0.0-rc.6",
     "body-parser": "^1.15.2",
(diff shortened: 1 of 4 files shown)


shubham-padia:
@Princu7 As `adm-zip` has been replaced, kindly please remove it from the npm dependencies, as it is no longer required.
Also noticed you are using an older version of the webapp, but that's not an issue as far as the fix is considered :smiley:  . 

 Thank You .

Princu7:
@shubham-padia Thanks for the suggestion!! :smile: Did it right away.

mariobehling:
Thank you. Let's see how this works in staging.

PR #894, opened 2016-11-04
Avoid internal scroll bars in small resolutions/mobile screens

### LINKED ISSUE #822: Avoid internal scroll bars in small resolutions/mobile screens
On the start page there is an internal scroll bar, which is ok in large resolutions but not suitable for mobile/small screens. Please see below.

Expected behaviour: Instead of scroll bars area is expanded.

screenshot from 2016-09-21 16 48 48

shubham-padia:
@mariobehling 
resolve #822 .

For demo visit https://shubham-padia.github.io/open-event-webapp/ .

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -294,6 +294,12 @@ a {
 }
 
 @media (max-width: 792px) {
+  .speaker-with-bio {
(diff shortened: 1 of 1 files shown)


mariobehling:
Excellent job!

PR #905, opened 2016-11-18
move generator require to the top

### LINKED ISSUE #904: generator require is not at the top
Over in ['app.js' file, I noticed that generator require was not done at the top](https://github.com/fossasia/open-event-webapp/blob/development/src/app.js#L59). I'd like to update it . Is this a welcomed contribution right now?

arrizalamin:
fixes #904

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@
     "rules": {},
     "env": {
         "es6": true,
-        "browser": true
+        "browser": true,
+        "node": true
     },
     "extends": "eslint:recommended"
 }
--- a/src/app.js
+++ b/src/app.js
@@ -6,6 +6,7 @@ const compression = require('compression');
 const bodyParser = require('body-parser');
 const path = require('path');
 var siofu = require("socketio-file-upload");
+var generator = require('./backend/generator.js');
 
 const config = require('../config.json');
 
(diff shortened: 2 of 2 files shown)


src/app.js
```diff
@@ -6,6 +6,7 @@ const compression = require('compression');
 const bodyParser = require('body-parser');
 const path = require('path');
 var siofu = require("socketio-file-upload");
+var generator = require('./backend/generator.js');
```
jayvdb: this is http://eslint.org/docs/rules/no-undef ?  maybe eslint needs to be configured..?

arrizalamin: yes, I think the [node environment](http://eslint.org/docs/user-guide/configuring) should be set to true.

jayvdb: Do it as part of this patch, or as a separate PR?

mariobehling: Patch it here if possible. Cheers.

.eslintrc
```diff
@@ -2,7 +2,8 @@
     "rules": {},
     "env": {
         "es6": true,
-        "browser": true
+        "browser": true,
+	"node": true
```
jayvdb: different indentation here.

jayvdb:
And please amend your current commit.
i.e. submit either one or two commits, not a third one to fix your own indentation bug.

Issue #899, opened 2016-11-13
Logical Error: Wrong approach in copying folder


aayusharora:
- The current generator will not generate app when number of files are less than expected i.e if images folder is not there or session.json is missing.
- This is because the current generator make a zip folder while copying the contents
- It removes it when the number of files copied is equal to a constant variable `filesToCopy` which is hardcoded as  7 here.
- This is wrong as the user can upload only event.json without speaker.json or vice-versa, In this case the app hangs with statement `Copying JSONS`.

The following code is responsible for this:

```
      var filesToCopy = 7;
      fs.readdir(appPath + '/zip' , function(err, list){
        var filesCopiedCounter = 0;
        
        if(err) {
          logger.addLog('Error', 'Error while reading directory', socket, err);
          return done(err);
        }

        function check(err) {

        if(err !== null) {
            return done(err);
          }

          filesCopiedCounter += 1;
          if(filesCopiedCounter === filesToCopy)
            fs.remove(appPath + '/zip', done);
        }

```

- The approach to be followed is to set a counter equal to the number of files in user's uploaded zip and decrease the counter whenever any file gets copied.
- If the value of counter becomes <=1, then remove the extra zip folder and continue.

Princu7:
@aayusharora Yes, Sorry for the mistake. In the original code, we were copying sessions, audio, images, event, tracks, speakers, sponsors, microlocations(whatever was available out of them) and not any other extra files which might be available in the zip folder. We can use the async library to accomplish this task easily without any need of counter variable. The code will be more robust and error prone this way. 
Here is the code

```
async.each(list, function(file, callback){
          var filePath = appPath + '/zip/' + file;

          function check(err){
            if(err !== null) {
              logger.addLog('Error', 'Error while copying folder', socket, err);
              callback(err);
            }
            else {
              callback(null);
            }
          }

          switch(file) {
            case 'audio':
            fs.copy(filePath, appPath + '/audio' , check); 
            break;
            case 'images':
            fs.copy(filePath, appPath + '/' + file, check); 
            break;
            case 'sessions':
            fs.copy(filePath, appPath + '/json/' + file, check); 
            break;
            case 'speakers':
            fs.copy(filePath, appPath + '/json/' + file, check); 
            break;
            case 'event':
            fs.copy(filePath, appPath + '/json/' + file, check);
            break;
            case 'tracks':
            fs.copy(filePath, appPath + '/json/' + file, check);
            break;
            case 'microlocations':
[...]

PR #908, opened 2016-11-19
Uses aync library for copying of the contents of the zip folder

Princu7:
Provides an alternative way to solve the issue #899.


mariobehling:
Ok, let's try this out and see how this goes for a while. Cheers.

Issue #910, opened 2016-12-01
Show the generated event folder without the necessary need to have the amazon s3 keys.


Princu7:
Currently, when the forks of the app are being deployed on Heroku, the event folder generated is not being shown because of the lack of the amazon S3 keys. Make changes in the **mailer.js** file of the project to make sure that the generated event folder is shown even if the S3 keys are not present.
![screenshot from 2016-12-01 09-45-03](https://cloud.githubusercontent.com/assets/8847265/20781772/ac5f724a-b7ab-11e6-97fd-1c6ef9a12c99.png)

PR #917, opened 2016-12-04
Fixes the bug which prevents the generation of event folder on local …

Princu7:
…machine which don't have Amazon S3 keys. Fixes the issue #910.
Now the forked versions of the webapp can generate event folders on their local machine.
screenshot from 2016-12-05 00-18-13


Princu7:
@mariobehling @aayusharora Please review. It is an urgent fix.

mariobehling:
Thanks!

Issue #927, opened 2016-12-25
Fixing the URL in the mail sent to the user.


mansikhemka:
The email sent to the user for the successful creation of the web app is incorrect. It lacks the %20 needed in place of spaces in the event name.

Therefore, the URL shows 404 on clicking.

(Based on the web app run on heroku.)

<img width="532" alt="screen shot 2016-12-26 at 12 17 50 am" src="https://cloud.githubusercontent.com/assets/13851011/21472450/c9ea60d2-cb00-11e6-986f-e84b028e3a7f.png">

Besides fixing this, I suggest that instead of having %20 in place of spaces in the event name in the URL, we can instead slugify it to make it less cumbersome.

Princu7:
@mansikhemka Thanks for reporting the issue. Putting dash or underscore in place of the spaces would be good in my opinion. Any suggestions?
Also would you like to work on this issue?

mansikhemka:
@Princu7 Definitely, I will place %20 instead of spaces.

fatimarafiqui:
@Princu7 Should we not use encodeURI? That will be more appropriate solution.

Princu7:
@fatimarafiqui Agree. That would be a better solution. Please open a pull request regarding it. Thanks

aayusharora:
Nice work @mansikhemka  Thanks :)

fatimarafiqui:
I will open the pull request shortly @Princu7 @aayusharora

PR #937, opened 2016-12-31
Used URI Component in the mail sent to the user

fatimarafiqui:
Used encodeURIComponent instead of replace function #927


mariobehling:
@fatimarafiqui Thanks for joining up.

It seems though like this issue is already resolved. Can you confirm, please?

fatimarafiqui:
Formerly the issue was resolved using replace function for just replacing all “%20” with a space in the URI. My solution makes use of the encodeURIComponent() function that encodes the given string. This function encodes special characters. In addition, it encodes the following characters as well: , / ? : @ & = + $ #. So in a way, It would be a permanent fix to all possible cases we may encounter in future.

mariobehling:
Oh nice. Yeah!

Thanks a lot. Merging it.

Issue #895, opened 2016-11-05
Implement ability to mark favorite sessions


alexagr:
Add ability for user to mark favorite sessions (e.g. with a star). The mark should be persistent across browser sessions (stored in local storage).

Princu7:
@alexagr I am up for the implementation. Will start working as soon as I get some spare time.

Princu7:
@mariobehling @aayusharora 
I have done a little work on this issue. I have added a star button to each session on the tracks page and the rooms page for quick bookmarking of the sessions. Below are the screenshots and the live link to the updated site
![screenshot from 2016-12-22 15-28-26](https://cloud.githubusercontent.com/assets/8847265/21421931/55b31c56-c85b-11e6-8126-db434538584b.png)
https://princu7.github.io/websiteTest

Now, we need to create a new page which will contain information about the bookmarked sessions. I need some suggestions and ideas regarding this new page. Like how much information to display about each session(room, time and type), displaying the name and picture of the speaker or not, giving detailed information about the session or not?

I think that just writing the name, time and speaker of the session would suffice. The user can then click on that session which will redirect him to that session on the tracks/rooms page.
Would love to hear your opinions on this. Thank you.

alexagr:
Very nice start! 
A few comments / suggestions: 
1. Right now only one session can be marked as favorite. It should be possible to mark multiple sessions.
2. Please add a toggle to the top of the page, allowing to show sessions marked as favorite only. IMO there is no need for a separate page.
3. We should have similar functionality on Schedule page as well.
4. Please consider using "3 star" score system - allowing user to mark session with 1, 2 or 3 stars. This may be used, for example, to differ between "nice to visit" and "absolute must" sessions.

Princu7:
Brilliant suggestions @alexgr. I will keep those in mind while implementing the feature. Thanks.

Princu7:
@mariobehling @aayusharora @alexagr So I did some more progress on this issue. I have implemented the bookmark feature on the tracks page(one star currently). Please review and test the feature here.

https://princu7.github.io/websiteTest/tracks.html

I also need some inputs on this. Where should we place the _Show Starred/Show All_ button and what should be its color? Also, regarding the 3-star system, what should be it's behavior? Should the default order between sessions be followed or should the sessions be displayed according to the number of stars inside a particular room/track. Like session having more number of stars is displayed on top and vice-versa.
Would love to hear your opinions and suggestions. Thanks.

[...]

PR #938, opened 2017-01-02
BookMark

Princu7:
@mariobehling @aayusharora 
Implements the bookmark feature discussed in the issue #895.  Please test and review the live website at the url: https://princu7.github.io/websiteTest
Thanks.


mariobehling:
@Princu7 Thank you. Very nice!

I wonder, if it would be possible with this method to have the stars on the different pages, e.g. a session that is starred on the tracks page, could also be starred on the schedule page.

mariobehling:
Merging it. Cheers.

Princu7:
@mariobehling I was actually wondering about the same issue and was about to ask the same. I do think that implementing it that way would be more intuitive and professional. I will surely work on it. Thanks.

PR #944, opened 2017-01-05
Change navbar color and opacity

hoangvanthien:
Issue #864 

screen shot 2017-01-05 at 15 45 43

### DIFF
--- a/src/backend/_scss/_base/_config.scss
+++ b/src/backend/_scss/_base/_config.scss
@@ -33,7 +33,7 @@ $event-border: #d2d6df;
 $overlayimage: rgba(0, 0, 0, 0.5);
 $border-event: rgba(0, 0, 0, 0.15);
 $border-venue: #e6e6e6;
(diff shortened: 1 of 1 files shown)


aayusharora:
LGTM @hoangvanthien  Thanks :)

mariobehling:
Sorry, but this is probably good for the mobile version, but not for the desktop version, where it changes the appearance of the site. The opacity is a feature of the desktop site, not a bug. Unfortunately I have to revert this as the desktop site does not fit the intended design outline anymore. It would be great, if you could make these changes for the mobile site only. Cheers.

hoangvanthien:
@mariobehling The point is that this PR was included in my two tasks on GCI which were accepted. Does reverting affect my contribution in the contest?

Princu7:
@hoangvanthien No, it shouldn't. You did it in good faith. But it would be infinitely better if you could help to make these changes for the mobile site only. Thanks!!

PR #950, opened 2017-01-06
Fixes #946 : loader resets if upload is cancelled.

### LINKED ISSUE #946: UX: Reset Loader when Zip is cancelled
Whenever the user click on cancel Uploading zip , the loader should reset.
## Steps to Reproduce
- Upload Zip and click generate app
- Cancel Zip when app is generated and upload new zip 

ui

geekyd:


Princu7:
Nicely Done. Thanks

Issue #953, opened 2017-01-06
Schedule information not shown


aayusharora:
## Steps to reproduce 

- Go to schedule page
- Click on any schedule
- Schedule div will not expand to provide schedule info.
Correct functionality on click:

![previous](https://cloud.githubusercontent.com/assets/12194719/21731632/480768d8-d47b-11e6-8b34-12d2dd7af314.png)

CosmicCoder96:
@aayusharora I would like to work on this.

Princu7:
@CosmicCoder96 Fixed. A javascript bug.

aayusharora:
Thanks @Princu7  :)

PR #955, opened 2017-01-07
Fixed schedule page

Princu7:
@aayusharora  @mariobehling Fixes the high priority issue #953. Sessions on schedule page were not opening on clicking them. Please review. Here is the live link
http://princu7.github.io/websiteTest


aayusharora:
Thanks!! @Princu7

PR #964, opened 2017-01-09
fixed #956 App crashes when invalid zip

### LINKED ISSUE #956: App crashes when invalid zip is uploaded
Web app crashes if the uploaded zip file is invalid (not contains required folders and files).

Steps to reproduce:

1. Go to app homepage
2. Upload any zip that does not contains /sessions or  /speakers or /event or other folders.
3. Click generate webapp
4. You can see app crashing in server logs.

error

I would like to work on this and submit a PR soon.
`

uttpal:
Fixes #956 app crashing if uploaded zip is invalid and not contains required json or files. 
Live Demo - https://openevent-app.herokuapp.com/

### DIFF
--- a/src/backend/generator.js
+++ b/src/backend/generator.js
@@ -76,24 +76,29 @@ function transformData(sessions, speakers, event, sponsors, tracksData, roomsDat
             roomsinfo, apptitle, speakerslist, timeList, metaauthor, tracknames
           });
         });
-        
+
       });
     });
   });
 }
 
 function getJsonData(reqOpts, next) {
-  const appFolder = reqOpts.email + '/' + fold.slugify(reqOpts.name);
-  const distJsonsPath = distHelper.distPath + '/' + appFolder + '/json';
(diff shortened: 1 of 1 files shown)


src/backend/generator.js
```diff
@@ -241,11 +246,20 @@ exports.createDistDir = function(req, socket, callback) {
       console.log('================================WRITING\n');
       if (emit) socket.emit('live.process', {donePercent: 70, status: "Compiling the HTML pages from templates" });
 
-      getJsonData(req.body, function(data) {
+      getJsonData(req.body, function(error, data) {
+        if (error) {
+          console.log('Error Invalid Zip');
```
poush: I think you can better add this message on a log file.

Princu7:
@uttpal Good work. Works great. Thanks!!

src/backend/generator.js
Princu7: @poush Thanks for the review. The error message has already been added to the error logs.

PR #967, opened 2017-01-10
Fixes #954 : Addition of a Navbar for tracks

### LINKED ISSUE #954: Implement  a nav bar for sessions/tracks
Currently there is no navbar on sessions/tracks pages and it becomes difficult to navigate through the sessions without a navbar. We could take inspiration from eventyay navbar from the link given below, on the webapp they are datewise so there would be an option to select the date and then the navbar would display the corresponding tracks according to the selected tab on navbar.

screen shot 2017-01-07 at 12 08 41 pm

Eventyay sessions page : https://eventyay.com/e/18252ab6/sessions/

PS: If the issue seems valid, I would like to work on this.

CosmicCoder96:
Here is the live preview of, the navbar on a site which gets generated:

https://cosmiccoder96.github.io/open-event-webapp/tracks.html

screen shot 2017-01-11 at 1 45 21 am

@aayusharora It doesn't cover much of the area , in smaller devices, rather is scrollable.

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1252,96 +1252,105 @@ a {
   top :0;
 }
 
-.c-tabs-nav {
+.tabs {
+  margin-left: -15px;
+}
+
+.tabs-nav {
   display: flex;
   list-style: none;
   margin: 0;
   padding: 0;
 }
 
-.c-tabs-nav__link {
+.tabs-nav-link {
+  background-color: $gray;
+  color: $white;
   flex: 1;
   margin-right: 4px;
   padding: 12px;
-  color: #fff;
-  background-color: #b3b3b3;
   text-align: center;
   transition: color 0.3s;
-}
 
-.c-tabs-nav__link:last-child {
-  margin-right: 0;
-}
+  &:last-child {
+    margin-right: 0;
+  }
 
-.c-tabs-nav__link:hover {
-  color: #6d6d6d;
-}
+  &:hover {
+    color: $border-venue;
+  }
 
-.c-tabs-nav__link.is-active {
-  color: #dc446e;
-  background-color: #e7e7e7;
-}
+  &.is-active {
+    background-color: $gray-extra-light;
+    color: $dark-black;
+  }
 
-.c-tabs-nav__link i,
-.c-tabs-nav__link span {
-  margin: 0;
-  padding: 0;
-  line-height: 1;
-}
+  i,
+  span {
+    line-height: 1;
+    margin: 0;
+    padding: 0;
+  }
 
-.c-tabs-nav__link i {
-  font-size: 18px;
-}
+  i {
+    font-size: 18px;
+  }
 
-.c-tabs-nav__link span {
-  display: none;
-  font-size: 18px;
+  span {
+    display: none;
+    font-size: 18px;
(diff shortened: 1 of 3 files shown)


src/backend/_scss/application.scss
```diff
@@ -1251,3 +1251,97 @@ a {
   right :0;
   top :0;
 }
+
+.c-tabs-nav {
```
aayusharora: @CosmicCoder96 
#### Better class names
make class name understandble

src/backend/_scss/application.scss
```diff
@@ -1251,3 +1251,97 @@ a {
...
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.c-tabs-nav__link {
+  flex: 1;
+  margin-right: 4px;
+  padding: 12px;
+  color: #fff;
```
aayusharora: ## Required
Use SASS nesting properties instead of using same class names .c-tabs-nav__link

src/backend/_scss/application.scss
```diff
@@ -1251,3 +1251,97 @@ a {
...
+  display: none;
+  background-color: #e7e7e7;
+}
+
+.c-tab.is-active {
+  display: block;
+
+}
+
+.c-tab__content {
```
aayusharora: ## Required
@CosmicCoder96  Dont use plain css under .scss files, please write proper SASS.

src/backend/_scss/application.scss
aayusharora: ## Required
Use all colors from config.scss instead.

src/backend/templates/tracks.hbs
```diff
@@ -24,11 +24,97 @@
   <link rel="stylesheet" href="./css/bootstrap.min.css">
   <link rel="stylesheet" href="./css/schedule.css">
 
+<script>(function() {
```
aayusharora: ## Required
- Make a seperate js file instead  of adding to track.hbs

src/backend/templates/tracks.hbs
```diff
@@ -24,11 +24,97 @@
...
+          var link = tabNavigationLinks[i];
+          handleClick(link, i);
+        }
+      }
+    };
+
+    /**
+     * handleClick
+     *
+     * @description Handles click event listeners on each of the links in the
```
aayusharora: ## Required
Why is this needed ? Please remove extra comments @CosmicCoder96

CosmicCoder96:
@aayusharora I have made the requested changes, Please review again.

mariobehling:
Thanks. Please use the color codes that exist on the page. Please do not use the pink color for links.

mariobehling:
Also, adjust the width to match the width of the following area.

CosmicCoder96:
@aayusharora @mariobehling Thank you for the review. I have made the necessary changes.

Princu7:
@CosmicCoder96 Great design and work. Just one small issue. On hovering over the dates, the text goes from white to black and since the background color is already black, the text becomes difficult to read. Please fix this. Thank you!!

CosmicCoder96:
@Princu7 I have changed the colour. Please have a look.

championswimmer:
Looks good now.

Issue #966, opened 2017-01-09
Web app does not works properly when multiple clients generate event sites within a short window of time,


Princu7:
Currently, the web app is not able to handle multiple requests around the same time. It behaves in a funny manner, delivering the wrong event site to the user. 

**Steps to reproduce**

* Open two tabs in the browser which will correspond to the two different users.
* Let User1 upload a sample zip. Don't press the Generate Button yet.
* Let User2 upload a sample zip. 
* Let User1 and User2  press the generate Button in any order.
* The event site generated for both the users will be exactly same even if they had uploaded a different zip file. To be more precise, the zip file which gets uploaded later will be converted into the event site.

It might be a bit difficult to try this so I made a video demonstrating this issue.  Here is the link
https://www.opentest.co/share/bc1a6f00d5d811e6a7cb6f2926412bab

@mariobehling @aayusharora This seems to be a major and interesting bug. I will work and try to solve this issue. Thanks.

aayusharora:
Thanks @Princu7. This is something that needs attention. This may be because of single folder that is being replaced again and again. This can cause problem at same time click. 
Do you like to take it ?

Princu7:
@aayusharora Yes. Thanks.

poush:
The one simple solution could be appending last 6-8 digits of timestamp to the file name. This part should be done by the server and on upload complete server can return the name of the file which will be sent to the server again when creating the web app.

aayusharora:
@Princu7 Will the client-{id} be different for the same user generating multiple apps at a single point of time.
Are they considered as seperate clients ?

aayusharora:
@poush the idea of timestamps looks good to me.

Princu7:
@aayusharora Yes. Whenever a connection to the server is made, an id is assigned to that socket. It is independent of the user and the time.
Here is the excerpt from app.js file
```
var id = 0;
io.on('connection', function(socket){
  id = id + 1;
  socket.userId = id; // we are using this as client-{id}
```

Since node has a single execution thread, we can guarantee that no two connections will ever be assigned the same id even if they make requests at the same time. There will be no race conditions.

Princu7:
Yes. Whenever a connection to the server is made, an id is assigned to that socket. It is independent of the user and the time.
Here is the excerpt from app.js file

var id = 0;
io.on('connection', function(socket){
  id = id + 1;
  socket.userId = id; // we are using this as client-{id}
})
[...]

PR #980, opened 2017-01-14
Fixed the bug when multiple clients generate event websites within a…

Princu7:
… short window of time. Resolves the Issue #966. 

I have assigned a number(ID) to every connection. This connection ID is independent of the user and time and is guaranteed to be unique for every connection due to the single threaded nature of the node.js. There were other good approaches suggested like using the concept of timestamp but they required additional works and safety checks so I went ahead with this minimalistic approach.

Here is a demo video when the same user generated two event websites
https://www.opentest.co/share/bd307220da2611e68531cf81a6426bba

Here is the link to the deployed app for testing
http://shrouded-ocean-89460.herokuapp.com/

@mariobehling @aayusharora Please review.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -13,6 +13,7 @@ const config = require('../config.json');
 var app = express();
 var server = require('http').Server(app);
 var io = require('socket.io')(server);
+var id = 0;
 
 
 app.use(compression());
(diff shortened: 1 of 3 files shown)


aayusharora:
Awesome Work @Princu7

Issue #993, opened 2017-01-17
App crashes in some cases


Princu7:
Sometimes, when the user generates the zip file, the app will crash with the error "Error while copying file.  No (insert_any_file_name) found." It is most probably due to the asynchronous nature of node.js which might have been accidently ignored in the previous commits.

Here are the error logs:
```
[ { [Error: ENOENT: no such file or directory, open '/home/princu/open-event-webapp/dist/a@a.com/tempproject1/zip/images/speakers/ViralB.Shah_96.jpg.new']
    errno: -2,
    code: 'ENOENT',
    syscall: 'open',
    path: '/home/princu/open-event-webapp/dist/a@a.com/tempproject1/zip/images/speakers/ViralB.Shah_96.jpg.new' },
  { [Error: ENOENT: no such file or directory, open '/home/princu/open-event-webapp/dist/a@a.com/tempproject1/zip/images/speakers/ZionNg_159.jpg.new']
    errno: -2,
    code: 'ENOENT',
    syscall: 'open',
    path: '/home/princu/open-event-webapp/dist/a@a.com/tempproject1/zip/images/speakers/ZionNg_159.jpg.new' } ]

```
![screenshot from 2017-01-17 18-35-45](https://cloud.githubusercontent.com/assets/8847265/22021526/e2c04e38-dce3-11e6-98c6-9ee43f81a788.png)

@aayusharora @mariobehling I am working on this and will try to fix it as soon as possible.

Abhi2424shek:
It's a file not found error basically if I am not wrong, can you give steps to reproduce the issue @princu7

Princu7:
@Abhi2424shek 
You just have to generate the event file by uploading the zip file. The error will throw up. I have almost solved the issue. It was due to the asynchronous nature of node.js. A function was reading a file after it had been deleted by another function. hence the file not found error.

PR #994, opened 2017-01-17
Fixes the issue #993 which made the app crash

Princu7:
Resolves the issue #993 . Fixed the crashing of the app which was happening due to the asynchronous nature of node.js. It was accidently not taken care of in the commit for changing the dimensions of the images of speakers and sponsors. Due to this, a file which had been deleted by a function was being read by another function which lead to the "File not Found" error.

@aayusharora @mariobehling Please test and review.
Here is the deployed app
http://shrouded-ocean-89460.herokuapp.com/

### DIFF
--- a/src/backend/dist.js
+++ b/src/backend/dist.js
@@ -235,20 +235,31 @@ module.exports = {
                 logger.addLog('Info', 'No sponsors images found', socket, err);
                 return 0;
               }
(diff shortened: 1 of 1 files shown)


Princu7:
@aayusharora Please review this request. Thanks!!

PR #995, opened 2017-01-17
Fixes #977: adds scroll to top button

### LINKED ISSUE #977: Add a minimalistic scroll to top sticky button on tracks page
When the the number of tracks and sessions is on the higher side , for instance on the fossasia'16 site generated with webapp, it becomes cumbersome to reach the top. If a user is in the midst of the page vertically, the only option is to scroll all the way to bottom and click 'back to top' link or manually scroll to the  top of the page. It can be avoided by implementing a very minimalistic scroll to top button which appears on the bottom right of the screen as soon as the user starts scrolling.
I can show the proposed UI, if it seems a valid issue.

CosmicCoder96:
It adds a minimalistic scroll to top button which appears only if the user starts scrolling.
Here is a preview: 
image

and here is a screenshot on mobile interfaces

image

Live preview is available at : 
https://cosmiccoder96.github.io/open-event-webapp/tracks.html

@aayusharora @Princu7 Please review

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,8 @@
     "env": {
         "es6": true,
         "browser": true,
-        "node": true
+        "node": true,
+        "jquery": true
     },
     "extends": "eslint:recommended"
 }
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1212,6 +1212,18 @@ a {
     font-family: sans-serif;
     margin: 0 auto;
   }
+  @media only screen and (max-width: 600px) {
+    float: none;
+
+    .fossasia-filter {
+      border-radius: 5px;
+      display: block;
+      float: none;
+      height: 32px;
+      margin: 0 auto;
+      width: 80%;
+    }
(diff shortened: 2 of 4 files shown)


Princu7:
@CosmicCoder96 Looks great. It would be great if you could increase the speed up a bit. It's seem a little slow when compared to the speed which we get on clicking the dates.

geekyd:
@CosmicCoder96 Great work!
I agree with @Princu7 it seems a little slow. Increasing the speed would be great.

CosmicCoder96:
@princu7 @geekyd  Thanks ! Have increased the scroll speed.

aayusharora:
@CosmicCoder96 Please squash the commits. Thanks !!

CosmicCoder96:
@aayusharora Thanks.I have squashed the commits.

CosmicCoder96:
@aayusharora I have squashed the commits.

Princu7:
@CosmicCoder96 Thanks.

Issue #988, opened 2017-01-15
UX : Improving the links leading to speakers page


geekyd:
When any link is clicked which leads to speakers page it does not auto scroll to the required speaker.
There is no automatic scroll to the required speaker like tracks page.

Please check the link
https://www.opentest.co/share/144e74e0dae911e68531cf81a6426bba

I will be working on this.

aayusharora:
cool ! go for this :+1:

geekyd:
@aayusharora Please review the app

Demo video:
https://www.useloom.com/share/c017fb40dd5311e6bb99ed2f6348f871

aayusharora:
@geekyd Cant access this. Please generate the PR to check.

geekyd:
@aayusharora sorry for the late response I was unwell , I have created a PR please review it.

geekyd:
@Princu7 @aayusharora I have solved this , please close the issue.

Princu7:
Thanks!!

PR #996, opened 2017-01-20
Improved links leading to speakers page #988

geekyd:
Adds auto scroll to the speaker for which the link was clicked.


aayusharora:
Thanks :)

PR #1006, opened 2017-01-28
merge dev to master

championswimmer:


championswimmer:
I don't think this can be merged this way. Has to be force-merged. There are 700+ commits now :/ 

What do you suggest @aayusharora @mariobehling

aayusharora:
Exactly !! @championswimmer  I agree. We can not merge this as the there are many commits.

Princu7:
@aayusharora @championswimmer The master branch will be more than one commits ahead of the development branch because of the recent commits of merging and reverting master. So, a force push might be a viable option.

PR #1008, opened 2017-01-31
fixes #1007 restricts only zip to be uploaded

### LINKED ISSUE #1007: Bug: Generator Logs and File Upload
- [ ] The user should not be able to upload any file if it is not `.zip` extension, show `message to upload zip extension`
- [ ] Clear all logs messages after clicking cross button
bug

geekyd:
@mariobehling @aayusharora @Princu7 
Please review the code.

Live demo at 

### DIFF
--- a/src/www/js/form.js
+++ b/src/www/js/form.js
@@ -21,6 +21,15 @@ $(document).ready(function () {
   uploader.resetFileInputs = false;
   uploader.listenOnInput(document.getElementById('siofu_input'));
 
+  uploader.addEventListener('choose', function(event) {
+      if($("#siofu_input").val().search('.zip') === -1) {
+        $("#siofu_input").val('');
+        statusText.css({'color' : 'red'});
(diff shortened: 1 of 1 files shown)


Princu7:
@geekyd Looks nice. It would be better if you could read about the `choose` event in the `socket-io-file- upload` documentation. It is fired right before the start event and we could verify the file name at that moment. If the file is not zip, then we can return false and no further events will be fired. This way, code related to the start event remains unchanged
What do you think? Would love your opinions.
https://www.npmjs.com/package/socketio-file-upload#choose
Also squash your commits into one. Thanks!!

geekyd:
@Princu7 Thats great idea. I read the doc and I'm implementing `choose` event.

Thank you for the feedback 😄

geekyd:
@aayusharora Please review this. :)

aayusharora:
Thanks you :)

Issue #952, opened 2017-01-06
UX: Stop website generation when zip is cancelled


geekyd:
The website generation does not stop even if the zip gets cancelled in between the generation process.

**Steps to reproduce the situation**

1. Upload the zip and hit generate app
2. Cancel the zip before the generation of the app is completed.

![ezgif com-gif-maker 1](https://cloud.githubusercontent.com/assets/12807846/21724106/ffc93ac8-d458-11e6-8f22-c74b5830222f.gif)

geekyd:
@aayusharora I have tried to solve the issue. please check the link and give your feedback.
https://open-event-web-appp.herokuapp.com/

Princu7:
@geekyd Looks great. Couple of points

* Please make the generation progress bar disappear on clicking the X button
* Clear up the existing logs.

aayusharora:
Looks good @Princu7 points seems valid in terms of UX. Better if we can achieve them with a transition .5 in hiding and showing the progress bar.

geekyd:
@Princu7 @aayusharora I have updated and added clearing the log feature.
But the progress does not stop as soon as the cancel but is clicked.

**Approach I have used**
- Created a variable in generator.js to maintain the status i.e if the user hits cancel or not
- Checked if the status is not false(cancel is not clicked) at start and end of each async.series function.

Can you please suggest if this is the correct approach?
Please give your feedback and suggest a solution.

Princu7:
@geekyd Thanks for the updates. I will look at the code and give feedback soon.

geekyd:
@Princu7 @aayusharora I suggest we should hide the generation progress bar if the upload gets cancelled and clear logs when the generation begins.
With hiding the bar we will stop the generation process as well this be a better user experience.

aayusharora:
@geekyd  Please take #946 also into consideration with your approaches. Both the issues will be solved by a single PR.

geekyd:
@aayusharora Yes it will solve both these issues with a single PR. Should I go ahead and implement the solution that i have suggested?

Princu7:
@geekyd Please hold on for a little while. I have reviewed your code. Excellent effort. Good work.  Some points

*  Be super careful while declaring global variables in the server-side code. Remeber, Nodejs is single threaded. All the connections execute on the same thread. You have made a variable 'appStatus' to maintain the status. By default, it is true and when you click on the X button, the value changes to false. You then check for it in each step of the generation process.

[...]

PR #1031, opened 2017-02-17
Fixed Cancel Generation Issue

SparshBansal:
#952 Solved

Now the website build is canceled on pressing the cancel button. There also should not be any problem with multiple users.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -63,10 +63,14 @@ io.on('connection', function(socket){
     if(socket.fileUploaded === false) {
       socket.fileAbort = true;
     }
+    generator.stopBuild(socket);
   });
 
   socket.on('live', function(formData) {
     var req = {body: formData};
+
+    generator.enableBuild(socket);
+
     generator.createDistDir(req, socket, function(appFolder, url) {
       socket.emit('live.ready', {
         appDir: appFolder,
--- a/src/backend/generator.js
+++ b/src/backend/generator.js
@@ -30,6 +30,8 @@ const roomstpl = handlebars.compile(fs.readFileSync(__dirname + '/templates/room
 const speakerstpl = handlebars.compile(fs.readFileSync(__dirname + '/templates/speakers.hbs').toString('utf-8'));
 const eventtpl = handlebars.compile(fs.readFileSync(__dirname + '/templates/event.hbs').toString('utf-8'));
 
+const statusMap = {};
+
 if (!String.linkify) {
   String.prototype.linkify = function() {
     var urlPattern = /\b(?:https?|ftp):\/\/[a-z0-9-+&@#\/%?=~_|!:,.;]*[a-z0-9-+&@#\/%=~_|]/gim;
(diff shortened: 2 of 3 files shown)


aayusharora:
Thanks @SparshBansal But can you please mention it is referred too.

SparshBansal:
@aayusharora  , I am sorry , i didn't get what you meant.

aayusharora:
@SparshBansal Please refer the Issue number in the description of this PR.

SparshBansal:
Issue #952 Solved

aayusharora:
Ok Cool Thanks

aayusharora:
@SparshBansal Can you please squash the commits.

Princu7:
@SparshBansal Hey. Thanks for the request. To make reviewing easier, could you please host the test application on Heroku, please?

SparshBansal:
@aayusharora , I have squashed the commits and will send a new PR

SparshBansal:
@Princu7 , I will host it on heroku for review

SparshBansal:
@Princu7 I have deployed it on heroku and here is the link : https://safe-sea-55983.herokuapp.com/

aayusharora:
Nice Work @SparshBansal

geekyd:
@SparshBansal Before sending a PR please let us know that you are working on the issue.
I checked the demo link, the generation does not stop immediately if X is clicked.
You can use a variable in socket object to keep the status.
Please look into it.

aayusharora:
@SparshBansal There are bugs in this PR. @geekyd Thanks for reporting.
Reverting this :(

SparshBansal:
@aayusharora
@geekyd 
It doesn't stop immediately because when the user issues a cancel request it is send to the back end server and if the server has started some asynchronous operation in the asynchronous series then I can't cancel it in between. I cancel it between 2 asynchronous steps. 
That's why it doesn't immediately cancel the build process. 
Also I notify the frontend client of cancellation only when it has actually cancelled it in the backend. That is between 2 asynchronous steps. 
I don't know how to cancel an asynchronous operation in between. Could you please guide me on that.

geekyd:
@SparshBansal We can hide the progress bar and the status bar as soon as the user hits cancel, and the rest will be handled by the backend.

As soon as x is clicked 
**Hide the progress bar and the status text**

As soon as the user uploads zip 
**show the progress bar and the status text**

@aayusharora Do you agree?

SparshBansal:
Yeah we can do that

SparshBansal:
@geekyd I think we also disable the upload zip element till the build gets cancelled in the backend and we receive an acknowledgement at the client side

SparshBansal:
@geekyd  Also I have noticed a bug while running the dev version hosted on heroku , If I cancelled uploading in the middle and retried again , it started from the same point it left off thus suggesting that the upload was not cancelled.

Princu7:
@SparshBansal The zip file upload is definitely canceled. It might be starting from middle maybe because of caching issue :)
To be more precise, it is because of this line
`uploader.resetFileInputs = false`
If you set it to true, it should be starting from beginning. 
Will check the test server soon. Net speed a bit slow now :smiley:

PR #1039, opened 2017-02-19
Add unit test to confirm the generation of the event site

Princu7:
Adds unit-test for confirmation whether the event site is being generated or not. In the past, there have been any instances where the tests pass successfully but the app generator stuck in the middle while generating the event site. This test aims to ensure that the whole event site is being generated. 

I will be adding more unit tests. This was a major critical test so added it first.
@aayusharora @mariobehling Please review. Thanks! :)


aayusharora:
Thanks @Princu7

Issue #1010, opened 2017-02-04
Feature : The generated web app should have share button


geekyd:
The webapp generated should have buttons to share the website links to social networking websites.

@mariobehling @Princu7 @aayusharora What are your thoughts on this?

aayusharora:
@geekyd Yeah it's great idea. Will you like to work for this ?

geekyd:
Yes i would love to work on this 😄 

What should we go with in your opinion ?
Should we add share links for all the sessions or a share button to share only event website link ?
I think we should add share button to all the sessions.

@aayusharora @Princu7 @mariobehling 
What do you think?

Princu7:
@geekyd Great idea. I think sharing link for the whole event rather than a particular session would be more better. Other event management systems also provide this option.

geekyd:
@Princu7 thanks for the feedback
@aayusharora What do you think?

aayusharora:
Exactly !!  sharing event sounds good :)

aayusharora:
@geekyd Can you update me on this :)

geekyd:
@aayusharora sorry about the delay, I had examination.
I will start working on this asap.

geekyd:
@aayusharora I have successfully added the share button for facebook and twitter. 
Check out 
https://open-event-web-appp.herokuapp.com/

I need your suggestion about where to place the social links.
One option is to provide the links on all the pages with a small slider present on every page for the social media websites.
something like this. 

![screen shot 2017-02-11 at 11 48 14 pm](https://cloud.githubusercontent.com/assets/12807846/22856303/a810f7f8-f0b4-11e6-8ad0-f7edf1b03711.png)

Princu7:
@geekyd Nice. Yeah. That is the standard norm.

aayusharora:
@geekyd We have a lot of space above search bar on right side. Can you try this mockup.
![12](https://cloud.githubusercontent.com/assets/12194719/23009388/bb959a4c-f43b-11e6-9ff7-4929d4473c1c.png)

geekyd:
@aayusharora Sure we can try that. 
I will create two mock-up designs and we can choose the better one, before PR.

aayusharora:
Will be awesome @geekyd !! Looking forward :)

geekyd:
@aayusharora 
Here are the two designs
**I think we should implement a side hover bar. What do you think?**

**Buttons**
[...]

PR #1041, opened 2017-02-19
Adds social share buttons to the web app #1010

geekyd:
Adds social sharing feature to the generated web app.

@mariobehling @aayusharora @Princu7 Please review.

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -659,10 +659,47 @@ a {
 
 .track-names {
   color: $black;
-  margin-top: 16%;
+  margin-top: 0;
+}
+
+.social-buttons {
+  float: right;
+  margin-bottom: 100px;
+  margin-right: 10px;
+  margin-top: 15px;
+}
+
+.social-button {
+  border-radius: 5px;
+  color: $white;
+  cursor: pointer;
+  display: inline-block;
+  height: 40px;
+  margin-right: 3px;
+  text-align: center;
+  width: 40px;
+
+  i {
+    height: 100%;
+    line-height: 40px;
+    width: 100%;
+  }
+}
+
+.facebook {
+  background-color: $facebook-icon;
+}
+
+.twitter {
+  background-color: $twitter-icon;
+}
+
+.google {
+  background-color: $google-icon;
 }
 
 @media (max-width: 990px) {
+
   .title-legend {
     float: left;
     width: 40%;
@@ -670,6 +707,30 @@ a {
 
   .track-names {
     margin-top: 0;
+  }
+
+  .social-buttons {
+    display: block;
+    float: left;
+    left: 0;
+    margin: 0;
+    position: fixed;
+    top: 150px;
+
+    .social-button {
+      border-radius: 0;
+      display: block;
+      height: 30px;
+      width: 30px;
+
+      i {
+        line-height: 30px;
+      }
+    }
+  }
+
+  .fa-2x {
+    transform: scale(0.65);
   }
 }
 
@@ -925,6 +986,36 @@ a {
 .middle {
   padding: 4%;
   text-align: center;
+}
+
+.social-container {
+  text-align: center;
+
+  .social-buttons {
+    float: none;
+    margin-bottom: 0;
+  }
+
+  h3 {
+    display: block;
+  }
+}
+
+@media (max-width: 990px) {
+  .social-container {
+    h3 {
+      display: none;
+    }
+  }
+
+  .title-legend {
+    float: left;
+    width: 40%;
+  }
+
+  .track-names {
+    margin-top: 0;
+  }
 }
 
 .sponsorscont {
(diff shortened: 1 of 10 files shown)


src/backend/_scss/application.scss
```diff
@@ -659,7 +659,79 @@ a {
...
+  .title-legend {
+    float: left;
+    width: 40%;
+  }
+
+  .track-names {
+    margin-top: 0;
+  }
+
+  .social-buttons {
```
aayusharora: @geekyd Not getting this. According to markup  `social-button class` should be nested  inside `social-buttons`

src/backend/_scss/application.scss
```diff
@@ -927,6 +999,25 @@ a {
...
+    margin-bottom: 0;
+  }
+
+  h3 {
+    display: block;
+  }
+}
+
+@media (max-width: 990px) {
+  .social-container h3 {
```
aayusharora: nest `h3` inside.

src/backend/_scss/application.scss
```diff
@@ -659,7 +659,79 @@ a {
...
+      line-height: 30px;
+    }
+  }
+
+  .fa-2x {
+    transform: scale(0.65);
+  }
 }
 
 @media (max-width: 990px) {
```
aayusharora: merge this media query with line `1015`

src/backend/assets/js/social.js
```diff
@@ -0,0 +1,17 @@
...
+
+  switch (source) {
+    case 'fb':
+      window.open('http://www.facebook.com/share.php?u=' + window.location, 'facebook', 'height=360, width=580');
+      break;
+    case 'tw':
+      window.open('http://twitter.com/home?status=' + window.location, 'twitter', 'height=360, width=580');
+      break;
+    case 'go':
+      window.open('https://plus.google.com/share?url=' + encodeURIComponent(window.location), 'google', 'height=360, width=580');
```
aayusharora: why do we need encodeURIcomponent here.

geekyd: Google plus was not working without encoding the url. So I added the it to encode the url.

src/backend/_scss/application.scss
```diff
@@ -659,10 +659,47 @@ a {
 
 .track-names {
   color: $black;
-  margin-top: 16%;
+  margin-top: 0;
+}
+
+.social-buttons {
```
aayusharora: @geekyd Why there is another `.social-buttons` class defined on line 712. Please remove this one.

src/backend/assets/js/social.js
```diff
@@ -0,0 +1,13 @@
...
+
+$('.social-button').click(function() {
+  var source = this.id.toString().split('-')[0];
+
+  if (source === 'fb') {
+    window.open('http://www.facebook.com/share.php?u=' + window.location, 'facebook', 'height=360, width=580');
+  } else if (source === 'tw') {
+    window.open('http://twitter.com/home?status=' + window.location, 'twitter', 'height=360, width=580');
+  } else if (source === 'go') {
+    window.open('https://plus.google.com/share?url=' + encodeURIComponent(window.location), 'google', 'height=360, width=580');
```
aayusharora: What is the use of `encodeURIComponent` here.

src/backend/_scss/application.scss
geekyd: .social-buttons on 712 is inside `@media`

src/backend/assets/js/social.js
geekyd: After testing the build on heruku, the url without `encodeURIComponent ` was invalid url for gplus, after adding `encodeURIComponent ` the url was no more invalid.

src/backend/_scss/application.scss
aayusharora: @geekyd, In that case make a mixin that can be @include in both of them.

src/backend/assets/js/social.js
aayusharora: @geekyd Check this too.

aayusharora: If it is due to special characters, it should be same for all fb and twitter links as well.

geekyd: @aayusharora I have noticed the url of the share window is auto encoded only for twitter, and not for facebook and google plus. So special characters might be buggy for unencoded urls.

src/backend/_scss/application.scss
geekyd: @aayusharora mixin is good idea but the tag at `665` and `712` are not repeated code.

src/backend/assets/js/social.js
geekyd: **Without Encoding**
screen shot 2017-02-20 at 3 05 00 pm


**With Encoding**

screen shot 2017-02-20 at 3 05 41 pm

aayusharora: ok Agree !!

mariobehling:
Good idea, but this implementation is not in line with the design of the site. I need to revert it therefore. Please propose a design implementation that is according to the existing design.

Issue #1104, opened 2017-03-01
Show also end times of sessions


mariobehling:
Originally we decided to only show start times of sessions, but after user feedback it seems better to also include the end time of sessions as people do not know how much "space time" is between sessions otherwise.

Therefore please implement to show session end times on all schedule, tracks, room pages.

![screenshot from 2017-03-02 00-41-35](https://cloud.githubusercontent.com/assets/1583873/23470395/3b2eae38-fee1-11e6-917b-da1b98d8cfdf.png)

geekyd:
@mariobehling I will make the changes, asap. Should i place the end time next to start time?

aayusharora:
@geekyd Please use the format startime - endtime

geekyd:
@aayusharora Im'm on it :) Thanks

geekyd:
@mariobehling @aayusharora The schedule page has all the events starting at same time grouped together, they might have different end time. Should we ungroup them and make the list?

Please guide. Thanks

aayusharora:
Hey @geekyd I guess there is a bug here too, if you see the timings for the two events on `17 March` it stated them separately `00:00`. If they have been grouped according to start time, they must be grouped together in the time without repetition of the timings.
Now, what I suggest is to ungroup them because the grouping seems already buggy and then too state the end time, be sure to sort them according to timings, same as it is done now.

mariobehling:
@geekyd Good point. Let's keep the schedule page as is for now and only change tracks and rooms page for the moment. Cheers.

geekyd:
@mariobehling We have made changes to the schedule page please check http://fossasia.github.io/open-event-webapp/schedule.html

Please give your feedback
Thanks.

Issue #1099, opened 2017-03-01
Substitute spaces with underscore characters to avoid broken URLs


mariobehling:
Substitute spaces with underscore characters to avoid broken URLs. 

Background: Direct links to tracks can contain spaces. Copying those URLs into other applications will result in broken URLs if the spaces are not counted as part of the URL, e.g. "http://2017.fossasia.org/tracks.html#2017-03-18-Hardware & Making".

geekyd:
I will have a look at it.

PR #1107, opened 2017-03-01
Fixes : #1104 & #1099 adds end time and improves URLs in tracks page

geekyd:
Solves #1104 
- Adds end time to all sessions in rooms, schedule and tracks page

- Improves direct links in tracks page 

Test server :

Screenshots :
screen shot 2017-03-02 at 1 47 40 am

### DIFF
--- a/src/backend/fold.js
+++ b/src/backend/fold.js
@@ -26,6 +26,10 @@ function slugify(str) {
   return str.replace(/[^\w]/g, '-').replace(/-+/g, '-').toLowerCase();
 }
 
+function replaceSpaceWithUnderscore(str) {
+  return str.replace(/ /g, '_');
+}
+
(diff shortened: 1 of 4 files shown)


aayusharora:
@geekyd Resolving #1099 partially. Copy the session link and paste it in browser, it will redirect to a bit incorrect location.

geekyd:
@aayusharora can you specify the page?

aayusharora:
@geekyd Even if you try the schedule page, it will scroll down but a bit above than it should be.

geekyd:
@aayusharora Okay got it, I think the other pages are good for the links.
Please do check them.

geekyd:
@aayusharora I think we should close #1099 as the problem statement of the issue is solved in tracks page.
I suggest that we should make a new issue for the schedule page, do a pr for it.

Issue #1105, opened 2017-03-01
Improving performance of Web app


aayusharora:
- Please see Google page insights and improve the performance of pages.
- [ ] Approaches we can adopt for RAIL model.
- [ ] Suggest and implement the ways to improve DOM and CSSOM construction, which is currently done at once.
- [ ] Remove render-blocking CSS
- [x] Add gulp to minify the JS, CSS code for production.
- [x] Compress images and scale them to 264px rather than 300px.
- [x] Add lazy loading for images

anu-007:
@aayusharora checked google page insights, I think we need to do more work on web app for mobile devices as the scores are low i.e. (63/100) and the performance on desktop is also average i.e. (78/100),I like to work on it.

Princu7:
I will read up on Gulp and try to integrate it. Will also compress the images using sharp module.

geekyd:
@aayusharora We also need to use the browser cache to cache the resources used by the app.

Princu7:
@geekyd Yeah, sure. Go ahead :+1:

geekyd:
@aayusharora I'm working on minification of the scripts and css files using gulp, will be sending a PR for it.

geekyd:
@aayusharora I have implemented the minification of all the css, js and html files using gulp.
Please check : https://open-event-web-appp.herokuapp.com
Also check the performance using https://developers.google.com/speed/pagespeed/insights/
The score has improved. Thanks

aayusharora:
@geekyd Thanks. Minification is working but I guess,  still there are 7 JS files which are requested as resources and same goes for 3 CSS files. When we give the app to production, we can send only 1 minified JS and CSS from Node server, Please make sure there will be no conflicts in the code when doing this. You can use IIFE in JS to remove same name variable/function conflicts.

geekyd:
@aayusharora So i should concat all the Js files and create a master script for web-app and same with Css??

aayusharora:
Hmm, let's leave CSS for now, planning to use `critical-css` module to only load necessary CSS on first load. JS we can minify and do.

geekyd:
@aayusharora Okay I'm on it. Thanks :)

aayusharora:
@geekyd Updates ?

geekyd:
@aayusharora I'm almost done with this, fixing the js files as they are causing problems after using iife and uglify. I will send a PR soon.

mariobehling:
[...]

PR #1132, opened 2017-03-06
Added lazy-loading for Images on speakers page

aayusharora:
Issue #1105 Loading Time Improvement
- Lazy loading Images in DOM result in significant Improvements of Web app
- On speaker page, the number of requests for Images reduces to 22 from 204.
- Only those images are requested which are in the viewport.
**Previous => loading Requests on first load**
previous
**After = >loading Requests on first load**
afterimprovement
Will be doing the same for other pages.
@mariobehling Please merge this.

### DIFF
--- a/src/backend/assets/js/jquery.lazyload.js
+++ b/src/backend/assets/js/jquery.lazyload.js
@@ -1,0 +1,242 @@
+/*!
+ * Lazy Load - jQuery plugin for lazy loading images
+ *
+ * Copyright (c) 2007-2015 Mika Tuupola
+ *
+ * Licensed under the MIT license:
+ *   http://www.opensource.org/licenses/mit-license.php
+ *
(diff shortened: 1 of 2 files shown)


championswimmer:
This is great. Would make page loading so much faster now. Thanks @aayusharora

mariobehling:
This is an issue: Before I was able to download the whole page at once and then walk around at locations in a conference where there is no Internet. Now I need to be online to see the whole page. Let's discuss this in more detail, please. We need to optimize the images, but in regards to lazy loading it does not help to achieve the goal of making this app work when people lost Internet connection (which is very common at  large events both for 4G and Wifi). Also see my comment here: https://github.com/fossasia/open-event-webapp/issues/1105#issuecomment-286018073

Sorry, I have to reverse that for now unfortunately.

PR #1139, opened 2017-03-08
Convert the speakers images into jpg format and reduce size

### LINKED ISSUE #1123: Speaker images saved as png and very large
The speakers images are saved as png and they are very large. Jpg is probably better and they should be much smaller.

Princu7:
Fixes #1123.
* The speaker images are now being converted to jpg instead of png.
* The size of the zip has reduced significantly by over 60%

Test Server:
http://secure-meadow-20680.herokuapp.com/

### DIFF
--- a/src/backend/dist.js
+++ b/src/backend/dist.js
@@ -96,6 +96,15 @@ const downloadJson = function(appPath, endpoint, jsonFile, cb) {
 
 };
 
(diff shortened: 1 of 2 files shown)


mariobehling:
Excellent job! This commit cut the size of the zip by 3 times. Hope we can get even smaller.

Issue #1145, opened 2017-03-10
Schedule: Make search bar full width on small screens


mariobehling:
The search bar has its own row on mobile screen. It should make use of the full 100% width on small screens.

![screenshot_2017-03-10-09-59-08](https://cloud.githubusercontent.com/assets/1583873/23779087/258e7128-0579-11e7-82f6-37cfe59a9903.png)

dinesh-sirvi:
will fix this...

sumedh123:
@dinesh-sirvi  Already opened a PR.

dinesh-sirvi:
no problem @sumedh123

PR #1147, opened 2017-03-10
#1145 : make search bar full width on small screens

sumedh123:
Solves: #1145

Made the search bar appear full width for small screens like mobile. This is how it looks now :-

fossasia


mariobehling:
In the description, please add in future Solves or Fixes follows by # and Issue Number. This will close the issue automatically. Thank you!

Issue #1146, opened 2017-03-10
Schedule: Buttons for days and "Show Starred" should have same size occupying 100% of width


mariobehling:
Days buttons and "Show Starred" are occupying one row, but they have different sizes. Expected: The buttons should have one size and possibly occupy 100% of the width.

![screenshot_2017-03-10-09-59-08](https://cloud.githubusercontent.com/assets/1583873/23779143/917a6be4-0579-11e7-9866-acdda3fc269d.png)

dinesh-sirvi:
I'm on it...

dinesh-sirvi:
Is this okay??
![screenshot_20170310-174723](https://cloud.githubusercontent.com/assets/10358362/23795203/47a45034-05ba-11e7-9e00-4e2443e6fac9.jpg)

Princu7:
Looks Great. Please make a PR. Thanks

Princu7:
Work done. Closing this

PR #1148, opened 2017-03-10
#1146 Schedule: Buttons for days and "Show Starred" should have same size occupying 100% of width

dinesh-sirvi:
Results:-
screenshot_20170310-174723
[https://dinesh-sirvi.github.io/open-webapp/](https://dinesh-sirvi.github.io/open-webapp/)


Princu7:
Looks Nice.

Issue #1173, opened 2017-03-14
Add a Home Button to the Deployment Page


Princu7:
On the Github deployment page, after the user has deployed the event site on his gh-pages, there is no button available on the page to go back to the generation page. The user has to manually type the  address or press the back button to go the home page of the app.

**Expected**
The deployment page should have a home button which takes the user back to the generation page.

sumedh123:
@Princu7  A button at bottom of the page?

Princu7:
@sumedh123 Yeah. Just add it wherever you think it would appear good with regards to the aesthetics of the page :)

sumedh123:
Okay @princu7

PR #1174, opened 2017-03-15
#1173 : add a link to website generator

sumedh123:
Fix for #1173 . Opinions? @Princu7

### DIFF
--- a/src/www/deploy.html
+++ b/src/www/deploy.html
@@ -18,13 +18,12 @@
     <div class = "container">
       <div class = "col-md-3"> </div>
       <div class = "col-md-6 col-xs-12" style="text-align:center">
-        <h3 id = "title"> Thanks for logging in to the github account. Press the button below to start the deployment process </h2>
-        <button id = "deploy" class = "btn btn-primary" > DEPLOY </button>
+        <h3 id = "title"> Thanks for logging in to the github account. Press the button below to start the deployment process </h3>
+        <button id = "deploy" class = "btn btn-primary" > DEPLOY </button><br><br>
         <button id = "abort" class = "btn btn-primary btn-danger" style="display:none" > ABORT </button>
         <div class="upload-progress progress" style = "display:none; height:20px; margin-top:20px; margin-bottom:20px;">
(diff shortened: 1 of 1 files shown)


Princu7:
@sumedh123 Thanks so much for the request :) I think that you misunderstood a bit. We only need to add a home button on the deployment page. No need to add it to the generated event sites. We want to add it on the Automatic deployment to Github page through which the user can go back to the event generator page.

Where we want to add the home button
webapp generator

Where we want to go after pressing the home button
webapp generator 1

sumedh123:
Ohh sorry I misunderstood. But what's the use of adding it to the deployment page?@princu7

sumedh123:
Okay it's a little change but it is useful.....I ll get it done

Princu7:
@sumedh123 No problem at all :) Yeah, it is just a small useful change. When the user has clicked on the deploy button and the whole deployment process has completed, it would be nice to have a button which can take the user back to the event generation page from which he/she originally came from. It kinda looks odd to leave the user stranded on the page with no links after the deployment process has completed :smile:

Princu7:
@sumedh123 Thanks for doing the required changes :) Code looks good. Two points:

* Please provide a screenshot of the new deploy page with the home button.
* Squash the commits into one.

sumedh123:
This is how it looks now.

src/www/js/deployProgress.js
```diff
@@ -1,4 +1,4 @@
-var socket = io.connect('https://opev-webgen-dev.herokuapp.com/deploy');
+var socket = io.connect('https://fossasia2017.herokuapp.com/deploy');
```
sumedh123: I will get it changed.

Princu7:
The screenshot looks really nice. Good job. Thanks for squashing the commits. But there is still one problem. Discard all the changes you made to the config.json file(Github client id, client secret, session secret and callback url) while testing the deployment part locally. Just retain the changes made in the deploy.html file in this pull request. Thanks :)

sumedh123:
@Princu7  Done.

Princu7:
Thanks!!

PR #1175, opened 2017-03-16
Adds improved lazy load to the webapp

geekyd:
Adds improved lazy load to the webapp. Previously the lazy load enabled the web app to load the images when they are present in the viewport. This PR improves the image process of the lazy load.

- Loads the image before they come in the viewport, loads the picture faster with scroll.
- Makes the app faster with lazy load.

@mariobehling @aayusharora @Princu7 Please review the performance of the web-app.

Demo link : 
https://geekyd.github.io/eventApplication

### DIFF
--- a/src/backend/assets/js/jquery.lazyload.js
+++ b/src/backend/assets/js/jquery.lazyload.js
@@ -1,0 +1,242 @@
+/*!
+ * Lazy Load - jQuery plugin for lazy loading images
+ *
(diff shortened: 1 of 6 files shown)


Princu7:
@geekyd Works much better than the previous version. However, if I scroll a little fast, then the problem reappears that the user can see the process of lazy loading of the images. But I guess that this is related to the issue of the speed of the network on which the user is present. The server is not able to load so many images at once and hence this can't be fully eradicated. Good work :)

PR #1195, opened 2017-03-20
Synchronizes bookmarks on all the pages

### LINKED ISSUE #1192: Bookmarked sessions not synchronized on all pages
**Actual Behaviour**

Currently, the bookmarked sessions are local to one page only. For example - a bookmarked sessions on the tracks page is not bookmarked on the schedule page.

**Expected Behaviour**

Since the sessions are same, it only makes sense that the effect of bookmarking a session on a particular page should affect the other pages also. They should work in a global scope with respect to the event rather than being localized to a single page. 

**Screenshots of the issue**
The bookmarked sessions are different on different pages.
`Schedule page`
screenshot from 2017-03-20 08-54-20
`Tracks page`
screenshot from 2017-03-20 08-54-36

Yes.

Princu7:
Fixes #1192 

**Changes**
Synchronizes the bookmarked sessions on all the pages of the event site. If the user bookmarks a session on the tracks page, then this change will be reflected on the other pages as well. The session bookmarked on a particular page will remain bookmarked on the other pages as well. The scope of the sessions starred  are global now.

**Demo-Link**
https://princu7.github.io/eventSiteWebApp/

### DIFF
--- a/src/backend/templates/rooms.hbs
+++ b/src/backend/templates/rooms.hbs
@@ -349,7 +349,7 @@
         });
 
         $('.room-filter').click(function(){
-          var temp = JSON.parse(localStorage['rooms']);
+          var temp = JSON.parse(localStorage['sessions']);
           var id = Number($(this).attr('id'));
           if (flag) {
             if (isAdded) {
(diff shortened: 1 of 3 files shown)


geekyd:
@Princu7 Looks nice to me. Great job!

PR #1223, opened 2017-04-17
Removes duplicate navbars for dates in tracks and rooms

### LINKED ISSUE #1206: Removing Redundancy from Track Page
- The dates are REDUNDANT written two times, once in the pills and then in the track filter bar.
- Remove the pills bar and filter content on the basis of track bar only.

Process 
-   Input: User clicks on 17 Jan on tracks bar.
-   Output: Tracks appear for the given dates and the page content is filtered according to date.

geekyd:
Fixes #1206 

Changes: Removes the duplicate navbars for dates in tracks and rooms page. Introduces a single unified navbar for dates and jump links in tracks and rooms page.
Screenshots for the change: 
**Before**
screen shot 2017-04-17 at 11 10 41 pm
screen shot 2017-04-17 at 11 11 22 pm

**After**
screen shot 2017-04-17 at 11 10 53 pm
screen shot 2017-04-17 at 11 11 13 pm

Demo server : https://open-event-web-appp.herokuapp.com
@mariobehling @aayusharora @Princu7 Please review. Thanks.

### DIFF
--- a/src/backend/assets/js/tabs.js
+++ b/src/backend/assets/js/tabs.js
@@ -2,7 +2,7 @@
 var tabs, i, myTabs;
 
 tabs = function(options) {
-  var goToTab, handleClick, init;
+  var goToTab, handleClick, init, handleDisplay;
   var el = document.querySelector(options.el);
   var tabNavigationLinks = el.querySelectorAll(options.tabNavigationLinks);
   var tabContentContainers = el.querySelectorAll(options.tabContentContainers);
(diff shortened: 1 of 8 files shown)


src/backend/assets/js/tabs.js
```diff
@@ -26,6 +30,19 @@ tabs = function(options) {
       goToTab(index);
     });
   };
+  handleDisplay = function(className) {
+    className = className.split(' ')[0];
```
aayusharora: This creates a global copy of className which can be accessed anywhere, `Use let className`

src/backend/assets/js/tabs.js
```diff
@@ -26,6 +30,19 @@ tabs = function(options) {
...
+    var dates = document.getElementsByClassName('date-filter');
+    for(i =0; i< dates.length; i++) {
+      if(dates[i].classList.contains(className)) {
```
aayusharora: Why should we write so much JS DOM optimization? Are we not using jQuery. Handling DOM with jQuery will be fast,
Please compare the performance  `document.getElementsByClassName` and `$('.classname')` using console.time()

src/backend/generator.js
```diff
@@ -61,6 +61,15 @@ handlebars.registerHelper('ifvalue', function (conditional, options) {
     }
 });
 
+handlebars.registerHelper('checkpage', function (conditional, options) {
+    if (conditional == options.hash.equals){
```
aayusharora: Please add some comments for this helper to understand this better
Also use === instead of `==`

mariobehling:
Please read my comment here: https://github.com/fossasia/open-event-webapp/issues/1206#issuecomment-295333262

This change does not follow design principles as far as I can see. The same buttons need to perform the same actions across one app. We are changing that.

PR #1246, opened 2017-04-28
Optimises calendar width for any number of room columns

### LINKED ISSUE #1233: Calendar View: Improve use of space
For calendar view:

1. Please make use of complete white area. Just keep a small space on left and right within the area to match the design.

2. Move sidebar with tracks below calendar in that view.

geekyd:
Fixes #1233 

Changes: Makes the width of the calendar adaptable. The calendar takes up 100% width of its container irrespective of the number of columns.

Screenshots for the change: 
**BEFORE**
screen shot 2017-04-27 at 7 38 05 pm

**AFTER**
screen shot 2017-04-28 at 12 40 07 pm

Demo server : https://open-event-web-appp.herokuapp.com
Gh-pages : https://geekyd.github.io/eventApplication

### DIFF
--- a/src/backend/templates/schedule.hbs
+++ b/src/backend/templates/schedule.hbs
@@ -349,6 +349,7 @@
                     </div>
                   </div>
                   {{/venue}}
+                  <div class="calendar-footer"></div>
                 </div>
               </div>
             </div>
@@ -528,6 +529,8 @@
           function adjustCalendar(element) {
             let rooms = $(element).find('.calendar-content').find('.rooms');
             let width = 0;
+            let offset = 0;
+            let roomCount = 0;
 
             $(rooms).find('.room').each(function() {
               width += $(this).outerWidth();
@@ -539,9 +542,18 @@
                     $(sessionName).html(textArray.join(' ') + '...');
                 }
               })
+              roomCount++;
             });
+
             if(width < $(rooms).width()) {
-              $(rooms).width(width);
+              offset = ($(rooms).width() - width)/roomCount;
+              $(rooms).find('.room').each(function() {
+                let currentWidth = $(this).width() + offset;
+                $(this).css({
+                  'width': currentWidth,
+                  'max-width': currentWidth
+                });
+              });
             }
           }
 
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1597,6 +1597,12 @@ a.skip:hover {
   display: block;
 }
 
+.calendar-footer {
+  position: absolute;
+  background-color: $white;
+  width: 100%;
+}
+
 .times {
   width: 48px;
   float: left;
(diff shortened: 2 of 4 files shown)


mariobehling:
We need to iterate more to ensure consistency across pages. Location of list/calendar view date heading is different to tracks/rooms page. Areas are not aligned in one left line. Indent is varying.

src/backend/templates/schedule.hbs
```diff
@@ -539,9 +542,18 @@
                     $(sessionName).html(textArray.join(' ') + '...');
                 }
               })
+              roomCount++;
             });
+
             if(width < $(rooms).width()) {
-              $(rooms).width(width);
+              offset = ($(rooms).width() - width)/roomCount;
```
aayusharora: Hey, @geekyd @Princu7 This is not the right approach to set width and current width every time the page loads with JS.
I hope you guys understand the effects of this on the performance of the app.
Proposed approach with some reasoning:
- Can we do this calculation on the server `calculation of offset` we already know the column width and we already know the number of columns. 
- After getting the `offset` as a handlebar variable, we can add it to HTML DOM element using CSS style. 
Please comment your suggestions @geekyd @Princu7.  :)

geekyd: @aayusharora Yes I agree with you

- We will need to store the width on the server, and we will stop using `min-width` & 'max-width' in css.
- We will calculate and set the width in handlebar while generating HTML.

aayusharora: Assuming we have a full-width container that will be 100vw ( 100% of the viewport ). 
I doubt why we need anything from frontend, we can structure the table like this.
- Calculate the width the table needs from server => It will be calculated as 
            container width = no. of columns * width  per column
Assuming we always want to give a table a full width.
width per-column = container width / no.of columns ( calculated at server )
This can be sent back to the client to take CSS.

I don't know if you are thinking complex cases for Responsiveness but we are not going into that for now.
Hope this helps :)

aayusharora: Also carrying the discussion further here https://github.com/fossasia/open-event-webapp/issues/1256

geekyd: @aayusharora We can go with what you are suggesting, But I'm not quite sure that would have any influence over the load speeds of the page. The function `adjustCalendar()` is only triggered when the user is in the calendar view and not `onLoad` of the page. If we really want to improve the load time of the page we should 

- Not execute the `display()` function onload. That would significantly help as on load we don't need to trigger it as it takes all the sessions of the page and checks if it should be displayed or not.

Please share your views on this.

aayusharora: @geekyd I always suggest measuring the improvements. 
Wherever a JS is involved making a CSS change, most importantly `Layouting` which is changing here, the whole cycle of Layout => Paint => Composite takes place which affects the performance.
Also if it contributes to performance increment of say 1ms ( which is very small ) but it will vary linearly with the data.   
Please carry the discussions on the issue.

PR #1257, opened 2017-05-01
Improves performance of calendar view

### LINKED ISSUE #1256: Performance Improvement in Table view
Hey, @geekyd @Princu7 This is not the right approach to set width and current width every time the page loads with JS.
I hope you guys understand the effects of this on the performance of the app.
Proposed approach with some reasoning:

Can we do this calculation on the server calculation of offset we already know the column width and we already know the number of columns?
After getting the offset as a handlebar variable, we can add it to HTML DOM element using CSS style.
Please comment your suggestions @geekyd @Princu7. :)

geekyd:
Fixes #1256 

Changes: Computes the width adjustment for calendar view while generation of the web app and removes the width adjustment code from schedule.hbs. 

Demo: 

### DIFF
--- a/src/backend/fold.js
+++ b/src/backend/fold.js
@@ -5,6 +5,8 @@ const distHelper = require('./dist');
 const urljoin = require('url-join');
 const async = require('async');
 const timeToPixel = 50; // 15 mins = 50 pixels
+const columnWidth = 160;
+const calendarWidth = 1060;
 
 function byProperty(key) {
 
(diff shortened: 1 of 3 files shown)


aayusharora:
@geekyd For a quick review, Can you please provide the gh-pages version to test.

src/backend/fold.js
```diff
@@ -91,6 +93,15 @@ function createTimeLine(startTime, endTime) {
   };
 }
 
+function checkWidth(columns) {
```
aayusharora: Nice !! :+1:

src/backend/templates/schedule.hbs
```diff
@@ -526,37 +523,6 @@
             $('.calendar-view').addClass('hide');
           });
 
-          function adjustCalendar(element) {
```
aayusharora: :1st_place_medal: For Getting rid of this. 
Looks correct.

aayusharora:
@geekyd The code is nice and solves a problem but I have to test a live version to get this Merged.
Please provide the live link.

geekyd:
@aayusharora Here is the link to gh-pages
https://geekyd.github.io/eventApplication/

mariobehling:
@geekyd Would you be interested to solve the same issue in the Open Event Orga Server? https://github.com/fossasia/open-event-orga-server/issues/3555

geekyd:
@mariobehling Sure I will try to solve it for the orga-server.

PR #1282, opened 2017-05-17
Set the font color of the tracks in accordance with the latest API spec

### LINKED ISSUE #1281: Adapt Font color depending on new API info
The font color is now set in the Orga server depending on the background color either black or white.

Please re-use the same color as in the schedule of the server. Follow up with the orga team about the API implementation and use color defined.

screenshot from 2017-05-16 15-42-18
screenshot from 2017-05-16 15-44-20

Princu7:
Fixes #1281 

**Changes**: The font color of the tracks on different pages(schedule, rooms, and tracks) are set in accordance with the latest API specification of the server. This will result in better UI and the font color will be more readable to the eyes and looks good on the background color.

* Schedule Page
screenshot from 2017-05-17 17-41-15

* Tracks Page
screenshot from 2017-05-17 17-41-29

**Live Link**: http://princu7.github.io/eventSiteWebApp

### DIFF
--- a/src/backend/fold.js
+++ b/src/backend/fold.js
@@ -40,6 +40,13 @@ function returnTrackColor(trackInfo, id) {
   return trackInfo[id];
 }
 
+function returnTrackFontColor(trackInfo, id) {
+  if ((trackInfo === null) || (id === null)) {
+    return '#000000';
+  }
+  return trackInfo[id];
+}
+
 function checkNullHtml(html) {
   html = html.replace(/<\/?[^>]+(>|$)/g, "").trim();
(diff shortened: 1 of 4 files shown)


src/backend/fold.js
```diff
@@ -40,6 +40,13 @@ function returnTrackColor(trackInfo, id) {
   return trackInfo[id];
 }
 
+function returnTrackFontColor(trackInfo, id) {
```
geekyd: Nice work!

mariobehling:
Good job!

PR #1285, opened 2017-05-19
Correct the behaviour of collapsible session elements on rooms page in

### LINKED ISSUE #1280: Collapsible elements behave differently on different browsers
I came across this bug while I was solving a previous issue. The bug can be best demonstrated with the help of some screenshots.

**On Chrome** (or any Chromium based browsers)

In the rooms page, if I click on any session, it expands correctly and then collapses correctly on clicking again. Here are the screenshots:

On Clicking the session element (In this case 'Opening Speech') 
screenshot from 2017-05-15 22-34-04

On Clicking Again to collapse
screenshot from 2017-05-15 22-34-08

**On Firefox**

On Clicking any session element(it expands normally) 
screenshot from 2017-05-15 22-37-19

**But it doesn't collapse correctly**. It creates an empty space. (only first session of the page collapse correctly)
screenshot from 2017-05-15 22-37-24

I am working on this issue.

Princu7:
Firefox browser.

Fixes #1280 

**Changes**: The session elements on the rooms page were not expanding and collapsible correctly. This was weird since the bug seems to only present on the Firefox browser and everything worked nicely on the Chrome. On further investigation, I found that the height of the session element was zero(because of floating elements inside it and hence it was collapsed). Due to this, Firefox was not able to correctly expand and collapse the element. Upon adding the property to make the height of the session element non-zero and computable, things started to work again.

* Before
screenshot from 2017-05-19 17-34-39
screenshot from 2017-05-19 17-34-45

* After
screenshot from 2017-05-19 17-34-52
screenshot from 2017-05-19 17-34-56

**Deployment Link**:  http://princu7.github.io/open-event-webapp

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1697,6 +1697,10 @@ a.skip:hover {
   position: relative;
 }
 
+.room-filter {
+  overflow: hidden;
+}
+
 .room-header {
   padding-left: 2px;
   padding-right: 2px;


mariobehling:
Excellent find!

PR #1296, opened 2017-05-27
Fixed the crashing of app when selecting Sengrid for sending emails

### LINKED ISSUE #1295: Using sendgrid for email crashes the application
**Actual Behaviour**

When using sendgrid instead of SMTP to send email, the application crashes. (but I still get the email)

**Expected Behaviour**

The app shouldn't crash and should send an email via the preferred option.

**Steps to reproduce it**

- In `config.json` change `DEFAULT_MAIL_STRATEGY` to lets say `sendgrid` (since `src/backend/mailer.js` seems to only check for `SMTP` and anything other than that results in using sendgrid)
- Add a valid sendgrid api key to `config.json` (As `SENDGRID_API_KEY`)
- Try to generate a webapp. 

**Log for the issue**

https://pastebin.com/ZhK45Bcg

**Screenshots of the issue**

N/A

No.

Princu7:
Fixes #1295 

**Changes**: 
* The problem was occurring due to the outdated version of the Sendgrid which didn't support Promises natively. Upped the version to the latest.
* Changed Sendgrid to the default option for sending email

screenshot from 2017-05-27 18-12-27
The Response code is 202

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 5.1.0
+  - 6.10.2
 sudo: required
 services:
   - docker
--- a/config.json
+++ b/config.json
@@ -1,6 +1,6 @@
 {
   "PORT": 5000,
-  "DEFAULT_MAIL_STRATEGY": "SMTP",
+  "DEFAULT_MAIL_STRATEGY": "Sendgrid",
   "DEFAULT_FROM_EMAIL": "info@eventyay.com",
   "SMTP_HOST": "mail.smtp2go.com",
   "SMTP_PORT": 2525,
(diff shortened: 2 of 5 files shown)


Princu7:
Not sure why the Travis failed. Looking into it. Please do not merge until then. Thanks!!

Princu7:
@niranjan94 Please review again. I had not updated the `shrinkwrap.json` file which stores the versions of all the dependencies while updating the Sendgrid package. Due to this, `npm install` which uses the shrinkwrap file was failing.  There were some more problems afterward, related to the use of `let` in the code. Fixed it by bumping the `node` version to 6.10.2  from  5.1. Now everything is working fine.

Issue #1217, opened 2017-04-08
Using Selenium for automated testing of generated event sites


Princu7:
`Selenium` is a browser automation library. Most often used for testing web applications, Selenium may be used for any task that requires automating interaction with the browser. In our case, it can be leveraged for automatic testing of the generated event sites. Currently, after every commit related to the improvement of event websites, we have to manually test the functioning and consistency of the site. We have to check for any regressions in the new version. Selenium will help us to eliminate this rudimentary step and prove to be helpful to both the contributors as well as the maintainers.

On top of my head, these are some of the areas where we can write tests using Selenium:
    
* Ticket section
* Search Bar
* Bookmark feature
* Date and Tracks Navigation Bar
* Different UI issues
* Dead Links
* Tweet section

There are different libraries for different languages for writing tests in Selenium. For `node`, we have the `selenium-webdriver` library. If it seems feasible, we can add the tests script to trigger on every request sent to the repo. It would help us to prevent regressions and maintain consistency. Would love to hear opinions, suggestions and even more tests. Thanks!!

geekyd:
We need to add selenium in the application. It will ease the testing process and we can do Dom component testing using the browser automation.

sumedh123:
@geekyd @Princu7 Actually yes. We have been into a situation where merging a PR broke a part of app. Then we had to correct it again to make it working. This is good idea. :+1:

Princu7:
@aayusharora Please provide feedback on this issue. In my humble opinion, we should maybe proceed with this issue first instead of working on other features. There have been many times where we had to revert a commit or make another request to fix the regressions introduced due to a faulty commit.  Using Selenium could help to catch some of those regressions earlier and reduce the manual labor. Please consider it. Thanks :)

aayusharora:
@Princu7 this can go parallely along with other issues. You can proceed with this, if you are willing to do so. Thanks

PR #1302, opened 2017-06-01
Add a very basic Selenium test to check the name of the event

Princu7:
The first step towards implementing Selenium testing in the project. Related to the parent issue #1217 

**Idea**
On every PR to the project, we will generate the event site (it can be anyone, currently I have set it to FOSSASIA 2016) and run a suite of functional tests against it. If the PR passes all the tests, it will pass else the Travis will fail.

**Changes**
* Run Selenium in Chromium browser (version 48) on Sauce Cloud.
* Add a very basic test to check the validity of the event name.

**Note**
* For better structure, readability, and scalability of the tests, we use [Selenium page objects model](https://watirmelon.blog/2015/10/30/webdriverjs-mocha-part-3-page-objects/). I will be working to incorporate that design into the tests and will soon open an issue consisting of a list of all the tests which we could run on the generated website.
* Also, we can change the browser to Firefox. Please guide on this.

**Test Link**
We can see the Travis logs to check whether the tests passed or not. Sauce Labs also provide us with a fantastic feature of watching the video of our tests. Since this is a very basic test, the video isn't that interesting and is very short. As we add more tests progressively, it will improve and its length will increase. 
https://saucelabs.com/beta/tests/7be5126fc8f147088cc9fcf061a9ebed/watch#4

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -17,6 +17,11 @@
       "from": "addressparser@1.0.1",
       "resolved": "https://registry.npmjs.org/addressparser/-/addressparser-1.0.1.tgz"
     },
+    "adm-zip": {
+      "version": "0.4.7",
+      "from": "adm-zip@>=0.4.7 <0.5.0",
+      "resolved": "http://registry.npmjs.org/adm-zip/-/adm-zip-0.4.7.tgz"
+    },
     "after": {
       "version": "0.8.2",
       "from": "after@0.8.2",
@@ -3190,6 +3195,11 @@
         }
       }
     },
+    "selenium-webdriver": {
+      "version": "3.4.0",
+      "from": "selenium-webdriver@latest",
+      "resolved": "http://registry.npmjs.org/selenium-webdriver/-/selenium-webdriver-3.4.0.tgz"
+    },
     "semver": {
       "version": "5.0.3",
       "from": "semver@>=5.0.1 <5.1.0",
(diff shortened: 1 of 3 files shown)


Princu7:
Looking at the Travis failures.

Princu7:
@mariobehling @aayusharora @geekyd  This is because of an issue with the Travis that for security reasons, it can't export the secure environment variables into the pull request builds. Here, Travis is unable to export the private `SAUCE_ACCESS_KEY` on Pull requests and hence is failing. 
https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

I am trying to find a workaround it. As an alternative, for the time being, we can publicly expose the variable and then when we find a solution regarding it, we can regenerate the access key and make it private again.

So, I tried changing the environment variables on Travis but for some reasons, they are having problems with the UI of the repository settings. And we are not able to access the settings page of the web app repository through which we can change the environment variables.
https://travis-ci.org/fossasia/open-event-webapp 

Here is an open issue for it:
https://github.com/travis-ci/travis-ci/issues/7831

They are saying that they have fixed the issue but for some repositories, the problem is still not solved. Sadly, our repository is one of them. And we have to send a mail to their support staff to rectify the problem. I will compose a mail and send to the Travis guys. Let's hope the problem gets fixed soon.
Also, we can still set the environment variables by directly editing the travis.yml file. Still, would be better to get this issue fixed nevertheless.

Princu7:
So I send an email to the Travis staff and the settings page works again now :sparkles:

PR #1319, opened 2017-06-06
Made Google's Open Sans font work offline

### LINKED ISSUE #1292: Avoid dependencies on external services
We have a dependency on the Google Api in the service without a real need, e.g. https://fonts.googleapis.com/css?family=Open+Sans:400,600,300 We could also use another font. The issue with this dependency is, that we are sharing our data without any need and if people are offline the browser tries to connect to the web for no apparent reason slowing down the page.

Please take out the dependency and implement another solution for the font.

Princu7:
Fixes #1292 

Changes:
* Makes the Google font Open Sans work offline now.
* Made appropriate changes to make sure all browsers are supported.
* Did the required settings with the help of this [site](https://google-webfonts-helper.herokuapp.com/fonts/open-sans). Tried many different variations and this one was the closest to the current one used on the site. The gulp library for adding the font during the generation process was not working properly.

**Deployment Link** : https://princu7.github.io/eventSiteWebApp/
Please Review. 
@aayusharora  @CosmicCoder96 @anu-007 @geekyd @harshitagupta30 @sumedh123 @utkarshnath.
Thanks!!


mariobehling:
Hit wrong button but merged now.

PR #1321, opened 2017-06-08
Enable single dedicated page for every session

### LINKED ISSUE #1314: Provide an option to show sessions on its own page instead of expanding it
This is a feature request, that could help us to get Mozilla using Open Event. So, there could be two ways to generate the web app:
a) A web app where the sessions expand as of now and
b) A web app where the sessions have their own page as in sched.org e.g. here: https://sanfranciscoallhandsjune2017.sched.com/

Please implement this in a way, that the user could choose this on the generators page through toggle or button.

Princu7:
Partially Fixes the issue #1314

Changes: 
* Provides an option for selecting a single dedicated page for each session or a common page for all the sessions.

**Deployment Link**
* Server: http://secure-meadow-20680.herokuapp.com


Princu7:
Two points are still not completed:
*  Make the rooms name into a link to the rooms list of the day
*  Make the tracks name into a link to the track of the day
There are some complications due to which these two couldn't be done in this PR. I will open an issue outlining the problems and possible solutions. But please review this. Thanks!!
@aayusharora @geekyd @mariobehling

aayusharora:
What is happening? Why codeclimate is falling on each PR @princu7 @geekyd ?

Princu7:
@aayusharora Don't really know. The added code in the PR will lead to a decrease in the code coverage of the entire project by some margin as visible in the CodeCov comment. Maybe because of it, it is failing.

@aayusharora @geekyd Updated the PR and the server link. Please review. Thanks!!

aayusharora:
You need to write tests then, these tests should always pass to ensure that the previous functionalities are working.
Please fix it so I can review this PR.
@mariobehling What's your suggestion in this ?

Princu7:
@aayusharora Added tests and resolved conflicts. Please review. Thanks

src/backend/assets/js/session.js
```diff
@@ -0,0 +1,66 @@
+$(document).ready(function() {
+  var sideslider = $('[data-toggle=collapse-side]');
+  var sel = sideslider.attr('data-target');
+  var top = $('header[role=banner]').outerHeight();
+  var downButton = $('#down-button');
+  var sessionElem = $('.single-session');
+  var sessionId = Number(sessionElem.attr('id'));
+  var maxNoOfSessions = 1000;
```
aayusharora: What is this limit for ?

Princu7: This just assumes that the maximum number of sessions in an event is 1000 and creates an array of that size. It is used for the bookmark feature.

aayusharora: What if it is more than thousand?

Princu7: Thousand seemed like a reasonable number so I set to it :grinning: . If it becomes more than that, then there shouldn't be any serious problem since the arrays are dynamic in nature in JS. We have also used the same number for tracks, rooms and schedule pages also. If we want to change it, we can perhaps make an issue and then fix it on all the pages at once. Please suggest. Thanks!!

aayusharora: Is that mean only 1000 sessions can be bookmarked? As, it is hard-coding the value in the localStorage.
Can we calculate length of the session object to set Maximum sessions?
It will save the memory and also we will not be going with a random value ?

Princu7: I thought about the current approach and it is true that it is a little quirky. It works correctly but it is not clean.

No. Any number of sessions can be bookmarked. We are not hard coding the values in the local Storage. Here is what we are doing presently:

* We check whether the `localStorage['sessions']` is defined or not. If the user is opening our app for the first time or has cleared his history, then that field will be undefined.

* If the field is not defined, we declare it and assign it a boolean array of 1000 indices. We now initialize the array to 0. The declaration size of the array and initialization of the array to zero is not necessary because of the dynamic nature of the Array.

* When the user clicks on a session, we find out its `sessionID` and mark its value as true on the same array index. That is, if the user clicks on the session which has an id of 5, then
`localStorage['session][5] = 1`

Now, this is where the little problem arises. The IDs are not sequential. They can be any number. Suppose if it is 2000, then although the array we have defined is of size 1000, `localStorage['sessions'][2000] = 1` will be done. To us, It will continue to work perfectly and we won't see any difference. But we are building up a sparse array behind the scenes(indices from 1001 to 1999 will be missing and value will be undefined) and although it won't be wrong, it will be hard to debug and is not a good code practice.

There might be multiple events so we are trying to make it a global to all. The session id of every session is unique regardless of the event. The best we can do here is to declare `localStorage['sessions']` as an `object` instead of an `array` and use the `id` of a `session` as a `field` in that `object`. That way, we would avoid creating a sparse array and write clean code.

Thanks for pointing out this problem.I think we should make a separate issue for it and solve it. This PR is already getting old and big. Please guide. Thanks!!

aayusharora: @Princu7 logically, session-id is a unique number that comes from API. So, it's maximum value should be the number of sessions logically.
Now, that length can be calculated on client side.
Also, creating key value pair ( object) dynamically whenever a session is bookmark ed seems reasonable rather than wasting memory in Array. Nice reasoning :)
Yes, make it a separate issue to progress.

aayusharora:
Can you pass code of tests for this PR, also update the description as 'partially fixes'.
Thanks @Princu7

Princu7:
I have removed most of the codacy issues. The rest of them don't make much sense. Please review. Thanks :)

aayusharora:
This contains my commit to in commit history @Princu7 `104550a`

Princu7:
@aayusharora Yep. Nothing bad. It's just that you updated the branch with the latest changes in the repo via the Github UI so it's showing that commit. Rebasing and sending again :)

aayusharora:
@geekyd  can u approve this, @Princu7 can u check the review of @geekyd 
It needs to be approved before merging.

Princu7:
@aayusharora Yep. Already made those changes along with the additional requirements. @geekyd Can you please review again? Thanks :)

geekyd:
@Princu7 Great Work :+1: !

PR #1335, opened 2017-06-12
Fixes the issue with the automatic deployment to Github feature.

### LINKED ISSUE #1334: Automatic Deployment to Github is not working currently
**Actual Behaviour**

Due to recent changes in the version of the packages used in the project, Automatic deployment to Github is not working currently. It just seems to complete but gets stuck on the last step of creating a gh-pages branch. This error is due to the recent change in the API spec of the library being used.

**Expected Behaviour**

Deployment should work properly and there should not be any problem.

**Steps to reproduce it**

Try to deploy any event to Github using the generator by clicking on the Deploy button.

Yes. Working on it. Thanks!!

Princu7:
Fixes #1334 

Changes:
* Made the automatic deploy to Github work again. The problem was arising due to the recent change in the API specification of a library being used in the project. Found and fixed it and the problem is solved.

**Server Link**: http://secure-meadow-20680.herokuapp.com

### DIFF
--- a/src/backend/deploy.js
+++ b/src/backend/deploy.js
@@ -174,7 +174,7 @@ module.exports = function(accessToken, folder, user, socket, callback) {
           return;
         }
 
-        var sha = res['object']['sha'];
+        var sha = res['data']['object']['sha'];
 
         gh.gitdata.createReference({owner: user, repo: repoName, ref: 'refs/heads/gh-pages', sha: sha}, function(err, res) {
           if(err) {


aayusharora:
Hi @Princu7 Can you please mention the link in the library specification where it is changed.
I have to see that, before merging this.
Also, please make sure to submit proper tests for each PR so that codecov  is not failed.
Thanks a lot :)

Princu7:
@aayusarora Thanks so much for the review :grinning: . Here is the link to the docs of the library which we are using:
http://mikedeboer.github.io/node-github/#api-gitdata-createReference

Mainly, the process related to the upload and commit of the files was working fine but the app was stuck when we were creating the GH-pages branch after the previous process has completed. For creating a branch from the latest commit, we have to take the `sha` reference from the current aka, the `master` branch here. The `sha` was not getting extracted properly. This is where the error was coming from. It's a bit hard to debug from the library because it doesn't mention the response clearly. You can also look at the Github API specs because the library which we are using is just a wrapper for that.

And the code coverage is not failing. It remains the same as the previous. It's just that the difference doesn't increase the coverage.

Princu7:
Okay, I found the documentation on the [Github official API docs](https://developer.github.com/v3/git/refs/#get-a-reference). This is what the response is supposed to look like:
```
{
  "ref": "refs/heads/featureA",
  "url": "https://api.github.com/repos/octocat/Hello-World/git/refs/heads/featureA",
  "object": {
    "type": "commit",
    "sha": "aa218f56b14c9653891f9e74264a383fa43fefbd",
    "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/aa218f56b14c9653891f9e74264a383fa43fefbd"
  }
}
```
The wrapper library which we are using provided us the exact same format in response earlier.
So, we extracted the `sha` by:
`var sha = res['object']['sha']`
Now the response after the recent version update of the library, the response was changed to something like this:

```
{
data: {
  "ref": "refs/heads/featureA",
  "url": "https://api.github.com/repos/octocat/Hello-World/git/refs/heads/featureA",
  "object": {
    "type": "commit",
    "sha": "aa218f56b14c9653891f9e74264a383fa43fefbd",
    "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/aa218f56b14c9653891f9e74264a383fa43fefbd"
   }
 }
}
```
So now we have to extract the `sha` like this:
`var sha = response['data']['object']['sha']`

Hope, it is a bit more clear now. Thanks :)

PR #1339, opened 2017-06-15
Scale the logo of generated events properly

### LINKED ISSUE #1277: Logo height not scaled resulting in increase top bar
The logo should be scaled to fit the top bar area. Currently it results in the expansion.
screenshot from 2017-05-15 12-03-05

Princu7:
Fixes #1277 

**Changes**:
* Scales the logo height of the events generated.

**Approach**
* There were alternatives ways to do this. One of them was to scale the logo to a fixed dimension during the generation process. The disadvantage of that approach was that the event logo comes in different size and shapes. So resizing them to a fixed size will change its aspect ratio and it will appear stretched and pixelated. A much simpler approach was to fix the max-height of the logo displayed on the site. The width of the logo will adjust automatically according to the height of the logo. This way, the height of the sub -navbar doesn't increase too much and the logo doesn't look stretched or pixelated.

Screenshots for the change: 
screenshot from 2017-06-15 16-37-11

**Test Server**:
http://peaceful-earth-19050.herokuapp.com

@aayusharora @geekyd Please review. Thanks!!


geekyd:
@Princu7 Have you tried this for other events?

Princu7:
@geekyd Yep. Generated app for the FOSSASIA Summit and it looked decent. Please test it. Thanks :)

src/backend/_scss/application.scss
```diff
@@ -129,7 +129,7 @@ a {
 
 /* Navbar adjustments */
 .logo {
-  max-height: 45px;
```
aayusharora: How will this ensure the aspect ratio @Princu7.

Princu7: I think that when both the max-height and max-width of the image are set, then the aspect ratio of the image is maintained. Please correct me if I am wrong. Thanks :)

Princu7:
@geekyd Thanks for the review. That is a better approach. Would change according to it :)

Princu7:
@geekyd @aayusharora I tried to follow the above approach. The issue is that the approach works well when the size of the logo is greater than 25px but when the size of the logo is below 25px, the UI suffers. eg: - on the FOSSASIA 16 event:
screenshot from 2017-06-15 22-45-27
Please suggest. What could be the correct approach? Resize the logo using the generator?

geekyd:
@Princu7 One solution could be scaling the logo while generating the web app. Here is how we can solve the problem:

- Get height & width of the logo in the API/ZIP.
- Calculate the aspect ratio width:height of the logo.
- Now scale it down while maintaining the aspect ratio.

By doing this we will have a common height & now we can use a fixed padding for the logo to place it on the navbar appropriately. @aayusharora Please share your views regarding this.

aayusharora:
@geekyd sounds correct, that's the only way to maintain this.
@Princu7 please follow this. Thanks :)

Princu7:
@geekyd Thanks for suggesting the way. I tried to do it that way. The small problem is that by transforming the size of the logo during the generation, we are able to resize it to a constant height, say, `45px`. And the width of the logo is then set according to the previous aspect ratio. Now in the event part, we have set the `max-width` of the logo to be `110px`. This means that whenever the width of the logo will cross `110px`, it will get trimmed and the height will automatically be set according to the `aspect ratio`.

This creates a problem in logo where the size of the width is much larger than `110px`. For Example,  `FOSSASIA 16`. Here the size of the logo after resizing comes around to be `264 * 45`. We have fixed height of `45px`. But since we have `max-width` of `110 px` applied to it, the height automatically adjusts to around `18px` in accordance with the `aspect ratio`. And with `padding` applied on the `anchor` tag above, it looks misaligned.

screenshot from 2017-06-16 09-26-18

One of the straight ways to solve the issue is to increase the `max-width` of the image. But in some cases, the width becomes a bit too large IMO. eg:- 
screenshot from 2017-06-16 09-27-43

I am trying to do some changes and see which looks best. Suggestions would be welcome. Thanks @aayusharora @geekyd

Princu7:
Ok. I tried many different settings and it seemed to work the best. Here are the screenshots of different events:-
* Facebook Developer Conference
screenshot from 2017-06-17 14-46-43
* Mozilla All Hands
screenshot from 2017-06-17 14-50-30
* Fossasia Summit
screenshot from 2017-06-17 14-51-40
* Open Tech Summit
screenshot from 2017-06-17 14-53-07
@aayusharora @geekyd Please review. Thanks!!

src/backend/_scss/application.scss
aayusharora: @Princu7 If the width of image is < max-width or height of image < max-height.
Max-width and Max-height will not be considered to calculate `aspect-ratio` as it is between width/height.

aayusharora: Yes, we do need to set max-width and max-height.
So, if you set an inline HTML `style` `` then HTML automatically maintains an aspect ratio of 1:1 to set the height, and we don't need to worry about this.
You can't hardcode width and height in CSS

Princu7:
@aayusharora
Sorry, this is taking longer than usual. Yes. Earlier, I was doing the same thing:

* Resized the logo image to a `fixed height` of `45px` and set its `width` according to the `initial aspect ratio` of the image. The height will `always` be 45px and only the `width` of the logo will `vary` after this process. Then applied a `padding` of `5px` on the `anchor` tag above the `image` tag.

The following property was applied on the image previously
`
max-height: 45px;
max-width: 110px;
`

The problem with it appeared when the width of the logo was much higher than `110px`. The exact size of FA logo after resizing was `264 * 45`. Now since the `max-width` set was `110px`, the height was automatically adjusted to `18px` (in accordance with the aspect ratio) and it looked like this:-

screenshot from 2017-06-16 09-26-18

To make it look better, we have to adjust the height properly **and** in order to do that,  we had to increase the `max-width` to around `220px` which was a bit too wide for a logo:-

screenshot from 2017-06-16 09-27-43

So, I went with the following setting, 
`
height: 45px;
max-width: 160px;
`
We know that the height of the logo will `surely` be `45px` and its width may vary (due to resizing step in the generation step). The height is hard-coded but it is because we are sure that the height of the logo will be `45px` due to resizing done in the generation step. The constraint is applied on the `max-width` of the image to ensure that it doesn't become `too large` in the case of some logos like FA. The logo of FA is still not perfect but looks better now (it has shrunk a bit due to the less width offered) and other logos are appearing perfectly. If we don't set height rigidly, then when the width is very large like in the case of FA logo, the height will shrink and it will become misaligned like shown in the previous screenshots.

I hope I was able to convey what I was thinking and doing. Thanks for the guidance. Really appreciate it. There might be a way to solve this using some javascript but please suggest any alternative suggestion. Thank you so much :)

geekyd:
@Princu7 Why do we need to apply `max-height` & `max-width` when we are resizing the image while generation?

Princu7:
@geekyd During generation, we are only fixing the height of the image to 45px and its width is set according to the initial aspect ratio of the image. Now, this width may be quite large and not suitable to display like in the FA logo, where the width comes out to be 264px. That's why I was applying only the `max-width` property not the `max-height` property since the height is fixed but the width may vary.

aayusharora:
@Princu7 Can you please check my review comment regarding this.

Princu7:
@aayusharora I am really sorry. Yes, I read that but didn't totally understand what you were saying. Can you please elaborate. Thank you so much :)

Princu7:
@aayusharora @geekyd So, I redid this using a different approach using dynamic padding of anchor tags above the images. The algorithm is:
* Declare variable padding = 5px
* Get the width, height and aspect ratio of the image.
* Set the height to 45px and calculate the width according to the aspect ratio. If the width <= 110px, then directly resize the image and no change in padding is necessary
* If the width > 110px, then make width constant to 110px and calculate height according to the aspect ratio. It will surely come less than 45px. Subtract the difference = (45 - height), divide it by 2 and add it to the padding variable.
* Apply padding variable on the anchor tag. Now every logo should be displayed nicely and we have fixed the height of the navigation bar = 55px for all cases.

Here are the screenshots:-
* FOSSASIA 2016
screenshot from 2017-06-18 20-45-11
* Mozilla All Hands
screenshot from 2017-06-18 20-46-03
* Facebook Developer Conference
screenshot from 2017-06-18 20-46-45
* Open Tech Summit 
screenshot from 2017-06-18 20-50-16

geekyd:
@aayusharora Please review this. It have been opened for a long time now. Thanks

Princu7:
Travis issues will get fixed after #1359.

Princu7:
Travis passing and fixed conflicts. @mariobehling @aayusharora @geekyd Please review and merge. Thanks :)

aayusharora:
@Princu7 This seems a nice algorithm, this can be more practical when you remove the hardcoded values 110px and 45px and calculate them using `jQuery`

aayusharora:
@Princu7  I am merging this right now, do the changes in next PR.

PR #1342, opened 2017-06-17
Enable Generation and hosting of different event sites along with an overview site listing them on Github Pages

### LINKED ISSUE #1320: Generate several test events on new PRs in gh-pages
At the moment we only generate on sample site to test the project. Event sites are different, have different length and number of tracks. So, reviewing different event sites continuously would be a good advantage.

Therefore please create an overview site at https://fossasia.github.io/open-event-webapp/ with links to event sites that should get created automatically on deployment:

Implement the following event sample sites from the samples in the Open Event repo:
* FOSSASIA 2017
* OpenTechSummit 2017
* Mozilla All Hands Hawaii
* F8 2017
* Google IO 2017 

While this may increase the build time, it is definitely a good way to continuously test and showcase the app.

Princu7:
Fixes #1320 

**Changes**:
* Generates different events viz FOSSASIA Summit, Open Tech Summit 2017, Mozilla All Hands and Facebook Developer Conference on each PR merge and host them along with an overview site listing them on the Github Pages.

**Test Link**: https://princu7.github.io/open-event-webapp

@aayusharora @geekyd Please Review. Thanks!!


test/serverTest.js
```diff
@@ -158,25 +158,111 @@ describe('app',  () =>  {
 });
 
 describe('generate', function() {
-  describe('.createDistDir()', function() {
-    this.timeout(200000);
+  describe('.create different event sites and copy assets of overview site', function() {
+    this.timeout(800000);
```
aayusharora: Thanks, Is this much of time required to generate the app.

gh_deploy.sh
```diff
@@ -10,19 +10,7 @@ fi
 
 rev=$(git rev-parse --short HEAD)
 
-HSAMPLE_EVENT="${HSAMPLE_EVENT:-Open\ Tech\ Summit}"
```
aayusharora: Why do we need to delete this code ?

test/serverTest.js
Princu7: Earlier, we were generating a single event and chose 2 minutes as the timeout limit. Now, we are generating 4 events so I thought it would be good if we give 8 minutes as the timeout limit. Practically, it doesn't take that much time for it to complete. This is just for when the Travis servers are working really slow and not downloading the assets fast enough. If it completes before this time, it will continue to the next test/finish so it should be fine. Thanks!!

gh_deploy.sh
Princu7: This was used previously for deploying Open Tech Summit 2016. Now, since we are already generating the Open Tech Summit 2017 Sample, there is no need of generating. Also, it is not mentioned in the list of sample events. So, I thought it would be better to remove it. Thanks!!

Issue #1345, opened 2017-06-17
Some bugs in webapp


mahikaw:
**Actual Behaviour**

- Event links on Instagram are not supported.

- The "position" of speakers should appear under the image of the speaker along with the organisation.

**Screenshots of the issue**

<img width="1238" alt="screen shot 2017-06-17 at 10 54 46 pm" src="https://user-images.githubusercontent.com/21010060/27255004-3ff97864-53b2-11e7-8373-260f3b6267b4.png">

Princu7:
@mahikaw The issues have been resolved. Please check and close it. Thanks :)

PR #1349, opened 2017-06-18
Added instagram icon as one of the event social link

Princu7:
Partially fixes the issue #1345 

**Changes**:
* Instagram link was not shown as one of the event links. Added it to the web app.

screenshot from 2017-06-18 17-11-14

There is an Instagram link at the top-right corner.

**Test Server**
https://secure-meadow-20680.herokuapp.com

### DIFF
--- a/src/backend/fold.js
+++ b/src/backend/fold.js
@@ -409,6 +409,9 @@ function createSocialLinks(event) {
       case 'google plus':
         link.icon = 'google-plus';
         break;
+      case 'instagram':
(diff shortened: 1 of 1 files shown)


Princu7:
@aayusharora The tests are failing because the sample used for testing currently is FOSSASIA 16 which didn't contain the Instagram link. No, the control never reaches the added lines of code. After the PR #1342, related to the generation of different events, this added part of the code will be covered and the code coverage will increase :)

PR #1355, opened 2017-06-20
Generate CNAME at time of deployment to Github Pages for redir…

Princu7:
…ection to custom domain

Changes: 
* The CNAME created in the GH-pages branch will get deleted every time when we deploy to the Github Pages through the existing script.  Added the generation of the CNAME file in the deploy script for proper redirection.

@niranjan94 @mariobehling @aayusharora Please review. Thanks :)

### DIFF
--- a/gh_deploy.sh
+++ b/gh_deploy.sh
@@ -32,6 +32,8 @@ git remote add upstream "https://$GH_TOKEN@github.com/"${TRAVIS_REPO_SLUG}".git"
 git fetch upstream
 git reset upstream/gh-pages
 
+echo $GH_CNAME > CNAME
+
 touch .
 
 git add -A .


gh_deploy.sh
```diff
@@ -32,6 +32,8 @@ git remote add upstream "https://$GH_TOKEN@github.com/"${TRAVIS_REPO_SLUG}".git"
 git fetch upstream
 git reset upstream/gh-pages
 
+echo 'https://sched.eventyay.com' > CNAME
```
niranjan94: Should be just `sched.eventyay.com`

niranjan94: Also, pick the value up from an environment variable called `GH_CNAME` instead. have added it to travis.

Princu7:
@niranjan94 Did the changes. Please check. Thanks :)

Issue #1373, opened 2017-06-27
Image jump and session element flicker and when expanded and collapsed


Princu7:
**Actual Behaviour**

Image jump and session element flicker and when expanded and collapsed. It's hard to describe it in words. So, here is the gif. It's a little hard to see here so look carefully.

![problem](https://user-images.githubusercontent.com/8847265/27573189-3717b700-5b2e-11e7-880a-96098e9eb809.gif)

This effect is visible on all the pages - schedule, tracks, and rooms.

**Expected Behaviour**

The session elements should expand and collapse smoothly with no hopping of images.

**Would you like to work on the issue?**

Yes. Working on it.Thanks!

PR #1374, opened 2017-06-27
Fix the image hopping/animation flicker issue on schedule page

Princu7:
Partially Fixes Issue #1373 

**Changes**:
* Eliminates image hopping/animation flicker when the session element is expanded and collapsed.
* Bootstrap collapse class is used for transition. It has a fix interval so the elements which are long will appear to have a faster animation (to fit in the time of transition) while the elements which are short will have a bit slower animation.

**GIF!**

fix

**Test Server**:-

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -902,9 +902,13 @@ a {
 }
 
 .track-inline {
-  margin : 1%;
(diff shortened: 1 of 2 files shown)


aayusharora:
Merging this, but needs some smooth transitions in that.

PR #1375, opened 2017-06-28
Implemented the bookmark feature in a better and efficient manner to …

### LINKED ISSUE #1314: Provide an option to show sessions on its own page instead of expanding it

Princu7:
…consume less storage space

Fixes #1314  completely. The details of the issue are discussed in detail here #1321.

Changes:
* Earlier, we were using an array for storing the ids of the bookmarked session which lead to a lot of wastage of space. So now using the object method for bookmarking of sessions for efficient utilization of memory.

**Test**:-
https://princu7.github.io/open-event-webapp

### DIFF
--- a/src/backend/assets/js/session.js
+++ b/src/backend/assets/js/session.js
@@ -1,20 +1,21 @@
 $(document).ready(function() {
+  // eventInfo is an object defined in the script section inside the session template file
+  var eventName = eventInfo.name;
   var sideslider = $('[data-toggle=collapse-side]');
   var sel = sideslider.attr('data-target');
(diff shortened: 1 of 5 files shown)


Princu7:
Done is #1380

PR #1378, opened 2017-06-28
Corrected track-list wrapping problem

### LINKED ISSUE #1356: Tracks list is wrapping in the layout-alignment issue
**Actual Behaviour**

Currently, when the number of sessions in a track are less- it looks like this.
screen shot 2017-06-19 at 11 58 33 am

**Expected Behaviour**

The tracks list shouldn't show up like this.

Yes. It will take sometime before I solve this as I'm working on other things as well, but I am willing to take up this issue..

mahikaw:
Fixes #1356
Changes: The track-list div was styled similar to content-block div so that it doesn't wrap around the parent container.

Screenshots for the change: 
screenshot from 2017-06-29 01-19-12

@aayusharora @Princu7 please review.

### DIFF
--- a/src/backend/templates/tracks.hbs
+++ b/src/backend/templates/tracks.hbs
@@ -243,7 +243,7 @@
               {{/days}}
             </div>
           </div>
-          <div class="track-names">
+          <div class="track-names col-md-9" >
             {{#tracknames}}
               {{#if title}}
                 <ul class="title-inline title-legend">


src/backend/templates/tracks.hbs
```diff
@@ -243,7 +243,7 @@
               {{/days}}
             </div>
           </div>
-          <div class="track-names">
+          <div class="col-md-9" >
```
aayusharora: You can not remove the track names class 
@mahikaw correct way is  `<div class="track-names col-md-9">`

mahikaw:
done @aayusharora Please review now.

Issue #1311, opened 2017-06-04
Test cases for Selenium testing of the website


Princu7:
These are the proposed test cases which we can check for fixing the regressions:

- [x]  Check whether the name of the event is correct or not
- [x]  Test the links of the navbar. The response code should not be 404.
- [x]  Test the links of the footer. The response code should not be 404.
- [x]  Search is working in both, Starred and All mode. Also, make sure that the search results change instantaneously on adding/removing a character.
- [x]  Tweets are being displayed on the event page.
- [ ]  Subnavbar on rooms and tracks page are working and we get appropriately scrolled down to the relevant section on clicking a link inside it.
- [x]   Clicking and expanding on the session element on all the pages work as expected. The element should collapse and uncollapse predictably.
- [x]  Share button is working.
- [x]  No scrollbars are present on the generated site.
- [x]  The down button which directs users back to the top of the page is working.
- [x]  After expanding the session element, when we click on the speaker picture, we should get redirected to the same speaker on the speaker's page.
- [x]  The track listed on hovering over the speaker's picture should redirect to the same track on the tracks page on clicking.
- [ ]  Check for the color of the font and the color of the tracks and make sure that it is correct.

I will start working on it and open child issues for them. Thanks!!

aayusharora:
Thanks for this. Nice sum up.

PR #1379, opened 2017-06-30
Added Selenium Test for checking tweet section on the page

Princu7:
Partially fixes the issue #1311 

**Changes**:
* The added tests check whether the tweet section is present or not.

**Video for Demonstration**
https://saucelabs.com/beta/tests/20dd281c00004f65b89bb053fea96755/watch#280
Not really visible since it is very fast but it is checking for the presence of the tweet section. If it was not there, the test would have failed.

@aayusharora @geekyd @sumedh123 Please review. Thanks :)

### DIFF
--- a/src/selenium/eventPage.js
+++ b/src/selenium/eventPage.js
@@ -10,4 +10,9 @@ EventPage.getEventName = function() {
   });
 };
 
+EventPage.checkTweetSection = function() {
+  //This will return an error if the tweet section is not there
+  return this.find(By.id('tweet'));
+};
+
 module.exports = EventPage;
--- a/test/serverTest.js
+++ b/test/serverTest.js
@@ -348,6 +348,14 @@ describe("Running Selenium tests on Chrome Driver", function() {
       });
     });
 
+    it('Checking the presence of tweet section', function(done) {
+      eventPage.checkTweetSection().then(function() {
+        done();
+      }).catch(function(err) {
+        done(err);
+      });
+    });
+
     it('Check whether the down button is working or not', function(done) {
       eventPage.checkDownButton().then(function(offset) {
         assert.equal(offset, 0);


Princu7:
Loklak seems to be down currently. That's why the tests are failing currently.

aayusharora:
Oh, can you please report this to loklak team on their channel @Princu7

Princu7:
@aayusharora Yes. I reported it. The server is up again but again the tweets are not coming in the response to the request made to the API. Loklak stores the tweets of last 30 days. In the past 30 days, there haven't been any direct tweet made by fossasia (There have been many retweets though!). So, It seems that the current API call doesn't support showing up of retweets in the response. I will confirm this on the channel. and follow up. Thanks :)

aayusharora:
https://loklak.org/api/search.json?timezoneOffset=-330&q=query This works perfect @Princu7

Princu7:
@aayusharora Both the queries work perfectly. But the problem is that they are not showing any retweets.
There haven't been any direct tweets from the fossasia twitter account in the last 30 days and only retweets are there. Since we are not able to fetch any retweets, the test is failing.

https://loklak.org/api/search.json?timezoneOffset=-330&q=from:fossasia
I have asked on the loklak channel about this issue. Will post here if get some info on it. Thanks!!

aayusharora:
Can you point me to the opened issue in loklak @Princu7  about the same.

Princu7:
@aayusharora Um. I had asked on the gitter channel since I thought I would receive a faster response there. Please see below
https://gitter.im/loklak/loklak
I will open an issue regarding it too. Thanks :)

Princu7:
@aayusharora Also, this tweet issue is taking some time to solve. We are generating many events right now like Mozilla All Hands 2017, Facebook Developer Conference, OSCON etc besides FOSSASIA Summit. Can we temporarily run the Selenium tests on another sample, say, FBF 2017 instead of the FA Summit? When the tweet issue is fixed, we can change it back to the FA Summit again. What do you suggest? Please guide. Thanks :)

Princu7:
Okay. I received a response from one of the developers of the loklak team and it seems that the retweets are currently not supported in the loklak server yet. Please see this link:-
https://github.com/loklak/loklak_server/issues/100
So what should we do? Changing the sample seem to be the only viable option to me. Please guide. Thanks :)

aayusharora:
I will suggest we should take a suggestion from @mariobehling in this case.

mariobehling:
I don't think including too many external services in the tests is a good idea.

Princu7:
Yes. Makes Sense. @aayusharora So what should we do. I think that instead of checking whether the actual tweets are being fetched or not. I guess it would be better if we just check that the tweet section is present in the site or not. That way, we are not depending on the external services and able to proide a basic test to ensure that the tweet functionality is present. Please guide. Thanks!!

arp95:
is loklak now showing tweets?
because i dont think they show up in any of samples in Open Event sample

Princu7:
@arp95 Loklak only stores the tweet of the last 30 days. Since almost of the Twitter accounts used in the sample have not made any direct tweets in the past month ( there have been some retweets from few accounts though but the loklak server doesn't support them presently), tweets are not showing up on the most of the event page.

Princu7:
@aayusharora This issue has been open for quite some time. Please guide on how to proceed with it. Please see my comment above. Thanks!! :)

Princu7:
I have modified the test to check whether the tweet section is present or not instead of checking whether the test section is present or not. If the tweet section is not present, then it will throw an error.  
@aayusharora @geekyd @sumedh123 @anu-007 Please review. Thanks :)

PR #1380, opened 2017-07-01
Made the animation on tracks pages smooth

Princu7:
**Partially Fixes** the parent issue #1374

**Changes**:
* Removes the image flickering issue on tracks page.
* Improves the animation and keep it in accordance with the tracks page

**GIF**: 
track

**Test Links**
https://princu7.github.io/open-event-webapp

**Please Merge after #1375**. Commit of that PR is included here because it was very difficult to separate them.

### DIFF
--- a/src/backend/assets/js/session.js
+++ b/src/backend/assets/js/session.js
@@ -1,20 +1,21 @@
 $(document).ready(function() {
+  // name is variable defined in the script in the session template file
+  var eventName = name;
   var sideslider = $('[data-toggle=collapse-side]');
   var sel = sideslider.attr('data-target');
   var top = $('header[role=banner]').outerHeight();
   var downButton = $('#down-button');
   var sessionElem = $('.single-session');
-  var sessionId = Number(sessionElem.attr('id'));
-  var maxNoOfSessions = 1000;
+  var sessionId = sessionElem.attr('id');
 
   function init() {
     var temp;
 
-    if (localStorage.hasOwnProperty('sessions') === false) {
-      localStorage.sessions = JSON.stringify(new Array(maxNoOfSessions).fill(0));
+    if (localStorage.hasOwnProperty(eventName) === false) {
+      localStorage[eventName] = '{}';
     }
 
-    temp = JSON.parse(localStorage.sessions);
+    temp = JSON.parse(localStorage[eventName]);
 
     if (temp[sessionId] === 1) {
       sessionElem.find('.bookmark').css('color', 'black');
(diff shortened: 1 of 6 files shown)


src/backend/assets/js/session.js
```diff
@@ -1,20 +1,21 @@
 $(document).ready(function() {
+  // name is variable defined in the script in the session template file
+  var eventName = name;
```
aayusharora: `name` should be in the local scope. How is this accessible here. It should be undefined here.

Princu7: It is accessible. There is some trick involved here. For individual session handlebar files, we have a separate external js file `session.js`. The name of the event is contained in the `jsonData` object passed to the `handlebars` file. Since the javascript file is externally linked to it, we can't directly use the `jsonData` object passed to the session handlebar file. So, we do something like this in the session handlebar file

```
<script>                                                                                                                       
      var name = "{{eventurls.name}}";                                                                                             
      name = name.substring(name.lastIndexOf('/') + 1);                                                                            </script> 
<script src="../js/session.min.js" type="text/javascript"></script>
```
By doing this, the variable `name` storing the name of the event get accessible to the external `session.js` file and we can use it there.

Princu7: This part is of the earlier commit of improving the bookmark implementation done in the PR #1375. Please merge this PR after that has been merged. I will have to rebase it again to remove the old commit. Thanks :)

aayusharora: Hmm. Thanks, is it good to keep name in the window then. Not sure, how can we share data between the js file with .hbs  without this.
But can we do it like `var a = {};
                                     a.name = {{eventurls.name}}

`

Now, `a` will be in window and not name. Than we access a.name in the other file.

Princu7: Thanks for the guidance :) Yes, I think that we have to keep a variable in the window. Otherwise, we won't be able to share the data between the hbs and js file. I will try using the approach you mentioned and update you. Thanks!

Princu7:
@aayusharora Made the appropriate changes here on this PR #1375. Here is the diff link 
https://github.com/fossasia/open-event-webapp/pull/1375/files
Created an object named `eventInfo` and made `name` as its property. Thanks for the guidance. Please review.

aayusharora:
Thanks @Princu7

Princu7:
Sorry, @aayusharora I had not included the object fix on this PR but on #1375 since it was part of that. But no worries. I will open a new PR for that. Thanks :)

aayusharora:
Sure.

PR #1382, opened 2017-07-03
fixes UI issues in sponsors and track-list

### LINKED ISSUE #1381: Recent sponosr and track list alignment changes are causing some UI issues
After the recent changes made, some UI issues have appeared.
**Old**
old

**New**
new

* The sponsor name should come underneath the image. But, it is appearing on the top of another sponsor which doesn't look good
* The text of the Map has become bold while it was normal font before
* The Track name list doesn't appear on the tracks page when we visit the site. It appears at the very end of the page.

@mahikaw Can you please look at these issues. Thanks :)

mahikaw:
Fixes #1381 

Changes: 

- Sponsor images and titles appear in a proper alignment now.

- The tracks list don't wrap or move to the end in any case(if the sessions are too less or more than the track-lists at the right)

Screenshots for the change: 

screen shot 2017-07-03 at 3 43 16 pm

screen shot 2017-07-03 at 3 49 23 pm

screen shot 2017-07-03 at 3 43 36 pm

----

@aayusharora @Princu7 Please review.

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -352,11 +352,13 @@ a {
 }
 
 .sponsor-row {
-  margin: 0;
   font-weight: bold;
   color: black;
-  margin-top: 100px;
-  margin-bottom: 100px;
+}
+
+.sponsor-row-text {
+  margin-bottom: 60px;
+  margin-top: 10px;
 }
 
 .sponsors {
@@ -371,6 +373,11 @@ a {
 
 .location {
   background-color: $white-classic-background;
+}
+
+.location-map {
+  margin-left: 0;
+  margin-right: 0
 }
 
 .address {
@@ -1823,11 +1830,6 @@ a.skip:hover {
   border-color:$light-black;
   margin: 0;
   width: 100%;
-}
-
-
-#content-block {
-  margin: 15px 0 0 15px;
 }
 
 #track-list {
--- a/src/backend/templates/event.hbs
+++ b/src/backend/templates/event.hbs
@@ -176,7 +176,7 @@
               <div class="row">
                 {{#each sponsorpics}}
                   {{#each this}}
-                    <div class="{{{divclass}}}">
+                    <div class="sponsor-row-text {{{divclass}}}">
                       <div class=" {{{sponsorimg}}} text-center">
                         <a href="{{{url}}}" data-toggle="tooltip" title="{{{type}}}">
                           <img class="lazy centre {{{imgsize}}}" alt="{{{name}}}" data-original="{{{logo}}}">
@@ -196,7 +196,7 @@
     {{#if eventurls.latitude }}
       {{#if eventurls.longitude}}
         <div class="location">
-          <div class="sponsor-row row">
+          <div class="location-map row">
             <div class="col-md-8" id="map">
             </div>
             <div class="address col-md-4">
--- a/src/backend/templates/tracks.hbs
+++ b/src/backend/templates/tracks.hbs
@@ -235,7 +235,7 @@
               {{/days}}
             </div>
           </div>
-          <div class="track-names col-md-9" >
+          <div class="track-names col-md-3" >
             {{#tracknames}}
               {{#if title}}
                 <ul class="title-inline title-legend">


Princu7:
@mahikaw Instead of adding extra classes and properties like `inline-block` and `float:right` which can cause some issue later on (`col-md-x` classes already use `float:left`), add `col-md-3` to the class of the track-names list div and remove the `margin` from the id `content-block` (because of it, the track-list names div shift at the very bottom instead of being adjacent to the session elements). It should work absolutely find then and there will be minimal changes code with more readability. Thanks :)

mahikaw:
@Princu7 That too seems to be a great way to do it but  I had thought that making separate classes for the tracks and sessions list will provide more flexibility to make individual changes in any of them separately, on the contrary. 
Also, here's the test link.https://mahikaw.github.io/testlink/  😄 
I have added an extra day in the event with just one session to recreate the scenario when the sessions are too less and the track list starts to wrap around the sessions list.

Princu7:
@mahikaw Thanks. It looks better now :) Yes, adding classes would allow greater customizability but they are not really required here. And they kinda conflict with the already used classes and the style of the bootstrap framework. Eg. on the track name list, we have used two classes: `col-md-3` and `content-block-track`. `col-md-3` already has `float: left` set internally but we are unnecessarily overriding it to `float:right` in the `content-block-track` class. It makes the code little confusing to read. Also, `display: inline-block` is not necessary when we are already using Bootstrap column classes.

In fact, if you completely remove the `content-block-track` and the `content-block-session` (while still keeping the `content-block` class removed), it would work absolutely fine :smile: So they are not really required here :)

Princu7:
Some more points:-
* There is too much space between the sponsors which makes the section very large. Can we reduce it please and constraint the size to what it was before? 
screenshot from 2017-07-04 10-52-34

* It also seems that unnecessary spaces have been added at many places on the front page :smile:

**Tweet Section**
* Before
screenshot from 2017-07-04 10-56-30
* After
screenshot from 2017-07-04 11-00-18

**Map Section**
Space has been added and the text has become bold now?
* Before
screenshot from 2017-07-04 10-56-45
* After
screenshot from 2017-07-04 10-55-47

mahikaw:
Okay @Princu7 I understand your point now. Just let @aayusharora confirm it once and then I can update it. Also, I deliberately made the text bold as you can see in the code changes. I feel it looks better this way. What do you think @Princu7 @aayusharora ?

Princu7:
@mahikaw Okay. Regarding the text being bold, I am not quite sure. It looks a bit out of the place on the page. Still, If we are going to keep the text bold, I think making the headline `Address` bold instead of the actual whole address bold would have been better but that is just my opinion :smile: @aayusharora Please guide on the on the different points covered so far in this PR. Thanks :)

aayusharora:
Hi, @Princu7 @mahikaw Previously we were on the stable version. Please make this as it was earlier.
Remove the extra spaces and make it less bolder.

mariobehling:
@mahikaw Thanks. Please only change what is mentioned in the issue. Let's avoid too long discussions.

@Princu7 Please help to guide @mahikaw more into the project. Thank you!

Princu7:
@mahikaw Please update the PR so it can be merged fast :smile: Thanks!

Princu7:
@aayusharora Also, please guide on the classes and approach to be used in the pull request. See the discussions above for more details regarding this. Thanks!!

mahikaw:
@aayusharora @Princu7 @mariobehling I have updated. Please review now. The test link is here  https://mahikaw.github.io/testlink/

Princu7:
@mahikaw Thank you so much. Many of the errors are fixed now. Sorry but there are still some problems left. Please see them:-

* Horizontal scrollbars are coming on the page. Please remove them!
screenshot from 2017-07-14 14-31-50

* There is a bit more space above and below the `Proudly Supported by` text in the speaker section. Just make the margin zero in the `sponsor-row` class. Also, add some padding below all the sponsor images. (in the `sponsorscont` class) It will look a little better I think!

* Current
screenshot from 2017-07-14 14-41-01
* After suggested changes
screenshot from 2017-07-14 14-42-54

Just do these changes and rest all is fine :smile: Thanks!!

Princu7:
@mahikaw Thanks for making the changes. It looks better now. You removed `sponsor-row` class becuase of which the div is overflowing and the horizontal scrollbars are appearing on the page. Just putting it back would remove them :plus1:

mahikaw:
@Princu7 Thanks for looking into it, but I tried doing that. The horizontal scrollbar still appears. Also, my reason behind removing the sponsor-row was that it has bold font, which makes the location text also bold. We don't want that, as discussed above.

Princu7:
Um. It is still appearing. Remember, we just need to make the margin of that row equal to zero. We can do that through any other class than doing it through `sponsor-row`. This is how it looks on my system
screenshot from 2017-07-14 16-15-13
Just use other class to set the left and right margin of the row to zero and it shall be fine :+1:

mahikaw:
It works now @Princu7 I have updated it. Thanks for your help! Please check now :)

PR #1389, opened 2017-07-10
Selenium test for checking the functionality of the down button

Princu7:
Partially Fixes the parent issue #1311 

Changes:
* Checks whether the down button which enables the user to go to the top of page on click is working fine or not.

**Sauce Video for Demonstration**
https://saucelabs.com/beta/tests/f69ef4b984f948fdad113a29af099854/watch#255

@aayusharora @geekyd @mahikaw @arp95 @sumedh123 @anu-007 Please review. Thanks!!

### DIFF
--- a/src/selenium/basePage.js
+++ b/src/selenium/basePage.js
@@ -65,20 +65,11 @@ var BasePage = {
   },
 
   checkDownButton: function() {
-
-    function scrollToEndOfPage() {
-      window.scrollTo(0, document.body.scrollHeight);
-    }
-
-    function getVerticalOffset() {
-      return window.scrollY;
-    }
-
     var self = this;
 
-    return self.driver.executeScript(scrollToEndOfPage).then(self.find.bind(self, By.id('down-button'))).then(function(el) {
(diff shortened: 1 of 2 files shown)


Princu7:
Don't know why the build has failed. Looking into it.

Princu7:
Something weird is going on. The tests are running locally but failing because of some problem with Sauce Labs. I kinda also faced this problem earlier. 
In WebdriverJS, you can execute a js script by specifying the function object or passing the string form of the function. When we pass the function object, it fails and when we pass the string form of the function , it passes. I opened a support ticket describing the problem in detail here:

https://support.saucelabs.com/hc/en-us/community/posts/115008390908-WebdriverJS-Execute-Script-not-working-properly

They then asked for including it in mail so I also sent a mail describing the problem to them. I am currently conversing with them and let's see if something comes up. Thanks!

Princu7:
Okay, so until the problem is sorted out, I have removed the function objects and used the string form of them in the tests.

mariobehling:
@Princu7 Please work actively with people to get two reviews.

PR #1401, opened 2017-07-16
Show complete dates

### LINKED ISSUE #1316: Show complete dates if space is available
Originally a short date form was chosen for buttons as the display was not providing enough space for long dates on mobile views. Showing short date formats e.g. months like "Mar" instead of "March" is a bit strange and there is no reason for this anymore as there are more elegant solutions.

Please propose and implement a better solution. Show complete dates on large screens and possible shorter date formats on smaller screens.
screenshot from 2017-06-06 00-24-13

mahikaw:
Fixes #1316 

Changes: As below.
Also, added back 'content-block' id which was removed by mistake in one of the commits.

Screenshots for the change: 

BEFORE:

screen shot 2017-07-16 at 3 36 30 pm

AFTER:

screen shot 2017-07-16 at 3 39 36 pm

------
I have checked, and it works on mobiles and smaller screens too. Here is the test link: https://mahikaw.github.io/testlink2/
@mariobehling @aayusharora @Princu7  Please review.

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1832,6 +1832,10 @@ a.skip:hover {
   width: 100%;
 }
 
+#content-block {
+   margin: 15px 0 0 15px;
+ }
+
 #track-list {
   margin-top:4%;
 }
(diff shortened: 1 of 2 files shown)


Princu7:
Hey. It is unrelated to this but a small problem has occurred after the latest changes. We unknowingly deleted the `content-block` attribute from the scss file instead of just removing it from a particular element. The result of it is that the date tabs and the headings below are not aligned anymore. Please correct it in this PR itself and modify the description a bit to reflect it. Thanks!!

mahikaw:
Done @Princu7 Please review now. :)

PR #1404, opened 2017-07-19
Check for scrollbars on index, tracks, schedule and room pages

Princu7:
Partially fixes the issue #1311 

**Changes**:
* Checks for the presence of scrollbars on all the pages of the site on small screen sizes. There is some work left. Speaker section is creating some problem. Also, have to add more sizes for check. Will complete it in another pull request.
* There were scroll bars on schedule page on small mobile size. Fixed it in this PR itself
* Before
screenshot from 2017-07-20 08-40-01
* After
screenshot from 2017-07-20 08-40-41

**Video for Demonstration**
https://saucelabs.com/beta/tests/b6676a7e436b467cb1e0f8cc86241bb8/watch#554

@aayusharora @geekyd @sumedh123 @arp95 @mahikaw @anu-007 Please review. Thanks :)

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -912,7 +912,11 @@ a {
   //cursor:pointer;
 }
 
-.track-inline {
+.schedule-track {
+  @media (max-width: 600px) {
+    display: block;
+  }
+
   margin-right: 10px;
   display: inline-block;
   vertical-align: text-top;
--- a/src/backend/templates/schedule.hbs
+++ b/src/backend/templates/schedule.hbs
@@ -107,7 +107,7 @@
 
               <div class="tracks col-xs-10 col-sm-10 col-md-10">
                 {{#sessions}}
-                <div class="track-inline" id="{{session_id}}">
+                <div class="schedule-track" id="{{session_id}}">
                   <a class="anchor" id="s-{{session_id}}"></a>
                   <div class="schedule-container">
                     <div class="padding-right-zero">
@@ -419,7 +419,7 @@
                   var temp = 1;
                   $(this).find('.time-filter').each(function() {
                       var flag = 1;
-                      $(this).find('.track-inline').each(function() {
+                      $(this).find('.schedule-track').each(function() {
                             if ($(this).find('.tip-description').text().toUpperCase().indexOf(filterVal.toUpperCase()) < 0 &&
                             $(this).find('.session-speakers-list a p span').text().toUpperCase().indexOf(filterVal.toUpperCase()) < 0 &&
                             $(this).find('.session-speakers-more').text().toUpperCase().indexOf(filterVal.toUpperCase()) < 0 &&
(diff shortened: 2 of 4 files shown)


Princu7:
@geekyd Thanks for the review. I have removed the conflicts and updated the Pull Request. 

@geekyd @sumedh123 There is one thing which I wanted to discuss. As you can see in the screenshot posted above, initially there were scrollbars on the schedule page on smaller screens. We were using `word-wrap:break-word` which split off the overflowing word to the oncoming line. But, in this case, there was a word `{distributed,remote,virtual}` which was quite large and hence couldn't fit onto the screen. To solve that problem, I added the `word-break: break-all` which break on every character. It solves the problem but it doesn't look too good as the word get cut in between on most of the sessions.

How do we proceed with it? One solution I think could maybe work is to check how big the scrollbar is. Like how much extra width is there which is causing the scrollbar to appear. If that extra width is not too much, we can they maybe pass the test. Please guide with this issue. Thanks :)

sumedh123:
@Princu7 I think the later one looks good. It's just that if we can get hyphens at the end words if they are breaking.

Princu7:
@sumedh123 Yeah seems a good idea. I will try to make changes for it. Thanks :)

Princu7:
@geekyd @sumedh123 I tried using the hyphens property but it only seems to work in the Firefox browser and not work in Chromium. Earlier, Chromium didn't support it but it seems to be supporting the property now and even though I have the latest version, I couldn't get it to work on Chromium. Please guide on it.

Also, what's weird is that the scrollbars are only introduced on the schedule page and not on the tracks page. The session which causes the overflow and introduces scrollbars in present on all the pages but for some reason, it is split on the tracks and rooms pages while it overflows on the schedule page. Here are some screenshots. Please guide. I think I am missing something silly but couldn't find it. Thanks :)

* Tracks Page
screenshot from 2017-07-23 18-36-06

* Schedule page
screenshot from 2017-07-23 18-43-43

geekyd:
@Princu7 Can you please provide a deployment link?

Princu7:
@geekyd Here is the deployment link
https://princu7.github.io/websiteTest/schedule.html
The hyphens property only work in Firefox and not in Chrome. Also, please read my previous comment. There are some other problems too. Please guide. Thanks!

geekyd:
@Princu7 The problem seems to be in `track-inline`, the display property is set to inline-block which is the reason why the word-wrap is not working for schedule page. 

What we can do is change the display to `block` for mobile screens & `inline-block` for other screens. It will help but will increase the height of page in some events. What do you think?

Princu7:
@geekyd Nice idea. Thanks! Yeah, seem to be the best solution right now to avoid this problem. The extra height won't cause much of a problem I guess. Will make changes in the PR incorporating this idea. For less than 600px, the display of the sessions will be `block` since up to that size, there won't be any difference between `inline-block` and `block` since most of the sessions will occupy the full width and for the screen size above that resolution, we will change it to `inline-block`.

geekyd:
@Princu7 Can you check if we can improve the quality of PR to pass codacy?

Princu7:
@geekyd Sure. I will look at them. Here are the updated screenshots :)
screenshot from 2017-07-29 18-40-03
screenshot from 2017-07-29 18-39-58

PR #1422, opened 2017-07-27
Fix the border issue in the calendar view of the session schedule

### LINKED ISSUE #1416: Use the same color and width of all borders
Borders style issue here is that it looks inconsistent. 
screenshot from 2017-07-24 10-00-24

Princu7:
Fixes #1416 

**Changes**:
* Fixes the border issue in the calendar view of the session schedule
* The border combined that's why the appearance was varied. The problem could be easily seen when we increase the border (eg: to 5px)

* 1px border
screenshot from 2017-07-27 18-13-30

* 5px border
screenshot from 2017-07-27 18-11-34
As we can see, when two sessions are together, the width combine resulting in weird appearance.

**After the Fix**
screenshot from 2017-07-27 18-13-49

@aayusharora @geekyd @sumedh123 @mahikaw @arp95 @anu-007 Please review. Thanks!

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1790,7 +1790,8 @@ a.skip:hover {
   @include cursorpointer;
   color: $white;
   position: absolute;
-  border: 1px solid $gray-extra-dark;
+  outline: 1px solid $gray-extra-dark;
+  padding: 5px;
   width: 100%;
   line-height: 120%;
   font-weight: bold;


Princu7:
Added padding of `5px` to the container. The text looks better now.
Here is the screenshot
screenshot from 2017-07-27 23-51-41
@mariobehling @geekyd @sumedh123 Please review. Thanks :)

Issue #1317, opened 2017-06-05
Implement the tracks filter using the existing right bar instead of top bar


mariobehling:
Right now we have a top bar with with tracks e.g. on the tracks page, but tracks and rooms page also have a right bar with all tracks that is currently not used. 

Please take out the tracks information from the top date bar and instead make the right side tracks to become a way to filter results. Please implement it in a similar way to the sched.org. Compare to the right side bar of https://hawaiiallhands2016.sched.com/ for example.

See also below (take out the light gray track info and use the sidebar info instead).
![screenshot from 2017-06-06 00-38-43](https://cloud.githubusercontent.com/assets/1583873/26806334/93ea1e8a-4a50-11e7-8829-2f3cd1b10734.png)

Princu7:
I have been thinking about how to implement the new design and there are some of the points I have come up with. On the bigger screens, there isn't much difference and the track name list present on the side can be used as a way to filter the results. On the smaller screens, the sessions occupy all the width and there is no space left to show the track name list which appears at the end of the page.

To solve this, we can make the list of the track names appear in the side bar navigation and the user can select any track from there. http://sched.org also implement it like this. Is this the correct approach. Please guide @aayusharora @niranjan94 @geekyd @mariobehling Thanks!

Princu7:
There are 4 major parts to be considered while doing this issue:

* The feature works on the desktop and tablet screen and the user can filter by clicking on the track from the side menu.
* On smaller screens, the track name list should appear in the side bar navigation menu like on the http://sched.org site and provide the same functionality.
* Now, we have many different filters which work on the session elements: - starred filter, search filter, and the new tracks filter. We have to ensure that all these filters not only work well in isolation but also when they are used in conjunction with the other filters. When implemented, this will be of great use to the user to quickly narrow down to the session which he/she is interested in.
* Write Selenium tests for this feature to ensure that future changes don't break the improvements made in this PR.

I will be working on each one of these points individually for easier development and review process. Thanks!

aayusharora:
Hi, @Princu7 Whenever you propose these kinds of design changes, it is always better to provide some mockups to clear your idea.

Princu7:
[...]

PR #1426, opened 2017-08-01
Make filters (track, search, and starred) work in conjuction with one another

Princu7:
Fixes the issue #1317 partially

**Changes**
* Make all the filters (track, search and starred) work in conjunction with one another.

**Test Server**

**Test Website**
http://princu7.github.io/websiteTest

@aayusharora @geekyd @sumedh123 @harshitagupta30 Please review. Thanks!

### DIFF
--- a/src/backend/templates/tracks.hbs
+++ b/src/backend/templates/tracks.hbs
@@ -269,6 +269,8 @@
       var flag = 0;
       var isAdded = 0;
       var mode = 0;
+      var trackFilterMode = 0;
+      var trackName;
       // Mode is true if the page is in starred view.
 
       $(function () {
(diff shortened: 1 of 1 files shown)


Princu7:
All the filters work well in isolation but it was a challenge to make all of them (track, search, and starred) work well in conjunction with one another. There are 3 filters and they can be applied in many orders (6 to be precise)

* Track -> Search -> Starred
* Track -> Starred -> Search
* Search -> Starred -> Track
* Search -> Track -> Starred
* Starred -> Search -> Track
* Starred -> Track -> Search

What's important to note is also that the reverse of all the filters applied should also work well. The same output which was before should display after reversing the filter too and no extra elements should be there.
eg:
* Track -> Search -> Starred -> Unstarred -> Unsearch -> Untrack

* Output after the 2nd and the 4th step should match
* Output after the 1st and the 5th step should match

I will be writing tests for making sure the improvements made in this PR don't break on future changes and send them in another PR. Thanks!

PR #1432, opened 2017-08-08
Added room name list for filtering sessions on smaller screens

### LINKED ISSUE #1318: Implement the rooms filter using a right bar instead of top bar on rooms page
Right now we have a top bar with with rooms e.g. on the rooms page. Please implement a rooms filter instead on the right hand side.

Take out the rooms information from the top date bar and instead make a right side tracks to become a way to filter results. Please implement it in a similar way to the sched.org but without the pop-up menu. Simply create a list of rooms here.

See also below (take out the light gray rooms info and create a sidebar info on the right above the track info instead).
screenshot from 2017-06-06 00-42-41

Princu7:
Fixes #1318 

Changes:
* Adds room name list filter on smaller screens.


Princu7:
**Test Website**
http://princu7.github.io/eventSiteWebApp

**Screenshot**
screenshot from 2017-08-09 20-37-20

@aayusharora @geekyd @sumedh123 @harshitagupta30 Please review. Thanks :)

Princu7:
Looking into why the tests are failing!

Princu7:
@geekyd Fixed the failing tests. Please review again. Thanks :)

Issue #1247, opened 2017-04-28
Implement offline features as PWA 


mariobehling:
Please implement offline features of app as a PWA as discussed in https://github.com/fossasia/open-event-webapp/issues/1228.

The PWA functionality should be implemented without any requirement of a server at this point of time.

anu-007:
@mariobehling  first of all, sorry for the late response as my exams are currently going on.

I like to add more on push notifications that we do not need any external database or server for notifications we can store relevant data in browser's indexedDB (like sessions alerts and bookmarked sessions information etc.) also there is no need of any server for this purpose we should use notifications which are pushed locally instead of pushed by server as in our web app we only want to display notifications about sessions currently going on and bookmarked sessions this is very basic task and I think it will be not appropriate to use a new server and database for doing only this task which can be performed locally using local notifications and indexedDB.
As we can extract the date and time to pop-up notifications from date and time as mentioned in sessions and the script will fire at that time to show that notification hence doing our work locally.
Although notifications pushed by the server has taken place of local notifications but in our case, i think we should stick to local notifications so I think we should tailor the features like notifications according to our needs instead of using conventional ones.Also, this will not have any effect related to the deployment of app anywhere like gh-pages etc. Thanks...!!!

mariobehling:
@anu-007 Thanks for the input. We can come back to it once we have a need for notifications. For now we mainly need offline features. So, let's close the discussion on PWA now and get back to the code. We need  offline features. Cheers.

anu-007:
@mariobehling sure we will first concentrate on offline feature first. Thanks...!!

aayusharora:
Any updates on this @anu-007 ?

anu-007:
@aayusharora I have not started to work on this issue as we have told to stabilize the web app first, but now I think it is right time to start implementing this feature, previously I have implemented some basic offline functionality to the web app which can be found here ( https://youtu.be/91uug0GuXhw ) so it does not take much time for me to implement it .I will start working on this and send a PR today.Thanks!!

Princu7:
@anu-007 Thanks. Please add a very basic service worker first. Basic installation and fewer cache contents. We would then need to improve and discuss the structure of the service worker and add more features to it. We would also need to create fallback pages when the cache retrieval fails. We would have to be slightly careful on removing and old worker and activating a new one.

anu-007:
@Princu7 ok I will take care of that.Thanks !!

Issue #1218, opened 2017-04-11
First step towards PWA: Installation, registration and activation of Service Workers


Princu7:
Based on the discussions in the issue #1211, we have decided to move forward in the direction of `progressive web apps` to make the event site more intuitive, responsive and pleasant to the user. `Service workers` are a key component of the `PWA`. They act like a proxy which sits between the client and the server, intercepting requests and can send custom made response back without the intervention of the server.

The first step to PWA is the successful installation and registration of service workers on the user's browser. There are different steps to this process. I have started working on this issue and will try to do the intended tasks soon. Thanks :)

PR #1437, opened 2017-08-14
Add service worker to the generated site for caching and offline loading of site

### LINKED ISSUE #1152: Adding a service worker for offline access of generated webapp
@mariobehling @championswimmer @aayusharora 
We should add a service worker to the generated webapp so as to-

* make the website function offline,
* increase online performance by reducing network requests for certain assets,
* provide a customized offline fallback experience.

Princu7:
Fixes #1152, #1218, #1247 

**Changes**
* Adds a service worker from caching and offline loading of pages. WHenever the content of the site is changed, the service worker will get updated again! This is done with the help of hash of the folder of the event website

**Test Server**
https://secure-meadow-20680.herokuapp.com

**Test Site**
https://princu7.github.io/websiteTest

@aayusharora @geekyd @sumedh123 @harshitagupta30 Please review. Thanks! More details are in the below comment!

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "decompress-zip": "^0.3.0",
     "express": "^4.13.4",
     "express-session": "^1.15.0",
+    "folder-hash": "^1.1.1",
     "fs": "0.0.2",
     "fs-extra": "^3.0.1",
     "ftp-deploy": "^1.1.0",
--- a/src/backend/assets/dependencies/offline.html
+++ b/src/backend/assets/dependencies/offline.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Page Title</title>
+    <style>
+      .error-template {
+        padding: 40px 15px;
+        text-align: center;
+      }
+      body {
+        background-color: #efefef;
+      }
+    </style>
+  </head>
+  <body>
+    <div class="container">
+      <div class="row">
+        <div class="col-md-12">
+          <div class="error-template">
+            <h1> Oops!</h1>
+            <h2> Page not cached </h2>
+            <div class="error-details">
+              You seem to be offline. Sorry, this page has not been cached yet. Other pages might be available!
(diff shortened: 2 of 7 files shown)


Princu7:
Ok, It looks complete I guess. 
What is done:-

* Whenever an event folder is generated, we calculate its hash and then use it as the cache key. This way, when we change the content of a site, the value of the hash changes, so the old service worker becomes obsolete and we see the new contents.

* All the static assets are cached when they are requested from the network so they are not fetched from the network again. Also, the pages which have already been cached will run comfortably on offline connections. In case a requested page which has not been cached is requested offline, there is a custom fallback page which displays the appropriate message.

Here are some of the screenshots
* When the site is updated, the previous cache contents are cleared
screenshot from 2017-08-15 13-35-17

* On visiting a page which has not yet been cached, its contents are placed into cache for utilization on future visits
screenshot from 2017-08-15 16-01-24

* On visting the cached page again, the assets will be served from the cache directly.
screenshot from 2017-08-15 13-36-48

* The page will comfortably load even on offline connections
screenshot from 2017-08-15 13-36-10

* On requesting for a page which has not yet been cached, we show a custom fallback page which shows a message
screenshot from 2017-08-15 13-36-22

Princu7:
**Some notes**

The code coverage is failing because few lines from the file `gulpfile.js` got commented accidentally in the previous commits. See them [here](https://gihub.com/fossasia/open-event-webapp/blob/development/src/backend/gulpfile.js#L20) So I uncommented it and because of it, the code coverage of that particular file went down (because that code which is related to failure of minifying process is not called during the generation process and hence on uncommenting that particular section of code, the code coverage went down a little). Other things seem to be fine.

I tried to remove most of the codacy and hound errors. Rest of them aren't much helpful. The variable `CACHE_NAME` is actually the hash of the event website folder which is injected into the file dynamically during the generation process itself after the folder has been generated and its hash has been calculated.

PR #1448, opened 2017-08-19
Integrates support for new open event server API in the webapp

### LINKED ISSUE #1322: Implement support for json-api
The Open Event Server is moving towards a new API as per the [json-api specification](http://jsonapi.org/). Hence all the client apps need to support the same. 

You can checkout some of the javascript client libraries that you can use to easily consume the API at http://jsonapi.org/implementations/#client-libraries-javascript. 

Please note that the zip format is not changing and only the API format is and the android app needs to support both. (Incase the user uploads a zip and generates an app.)

Princu7:
Fixes #1322 

Changes:
* Integrates support for new open event server API in the webapp


Princu7:
@niranjan94 @aayusharora @geekyd This integrates the new API support in the webapp. I created a test event using the new open event server API server. Please check it here
http://open-event-api.herokuapp.com/v1/events/297
The test server is http://shrouded-ocean-89460.herokuapp.com. 
It works on the above sample event and generates a small website for it. The problem is that it doesn't work with the old API because of the different attributes and the type of response. What can we do here? Add support for old attributes? Please guide. Thanks :)

niranjan94:
@Princu7 we will not be supporting old attributes

Princu7:
@niranjan94 Okay. Thanks! Really sorry but I am not really clear on how to proceed here. Do I put this PR on hold until the whole migration takes place? All the sample events that we display on the showcase site (From the Github and the old eventyay server) have the old attributes. So, we would also have to change them for the showcase site to work. Please guide. Thanks :)

Princu7:
@niranjan94 @aayusharora Really really sorry for pestering but could you please guide on how to proceed with this PR? I think we can put it on hold until the complete migration is not completed. What do you suggest? Please guide. Thanks :)

niranjan94:
@Princu7 if I remember correctly, there was a python script that handled changing the format of the old zips to the new. Check with the android guys. To see how they work with the samples.

Princu7:
@niranjan94 Thanks. Yes, there is a [python script](https://github.com/arp95/script_for_samples) which converts the old attributes to the new attributes in the samples but it is not used on any of the samples yet. 

I talked with Shailesh regarding this and he said that initially they had added support for both, new and old attributes, but later removed the support for the old attributes. The attributes are processed by the parser at the run time so if we generate the apk for the old sample, it will generate successfully but will throw an error when the user will open it. 

In webapp generator, the attributes are parsed and processed at the generation time itself. So if we remove support for the old attributes, then the generation of the event site will fail. The possible options are

* Add support for the old attributes too
* Update the samples present on the open event repo with the new attributes.

Please guide. Thanks :)

Princu7:
@niranjan94 @aayusharora Most of the samples in the open-event repo have been updated. This PR migrates to the new API and model of the open event server. Travis is green again. Sample events will be generated after each merge to the repo now. After this, the events on the old eventyay server will stop working completely.
The auto-deployment functionality works again

**Test Website**
https://princu7.github.io/open-event-website

Please review. Thanks :)

src/backend/dist.js
```diff
@@ -67,42 +68,66 @@ const downloadAudioFile = function(url, filePath, next) {
...
+    }
+    return cb();
+  });
 
   try {
-    console.log('Downloading ' + endpoint + '/' + jsonFile);
+    console.log('Downloading ' + endpoint);
     request
-      .get({url:endpoint + '/' + jsonFile,timeout:30000})
+      .get({url:endpoint, timeout:30000})
```
aayusharora: on what parameters, this value is set to 30000

src/backend/dist.js
```diff
@@ -522,15 +547,44 @@ module.exports = {
...
+    Object.keys(jsonsUrl).forEach(function(key) {
+      jsonsUrl[key] = endpoint + '/' + key;
+    });
+
+    if(endpoint.search('eventyay') != -1) {
+      endpointType = 'eventyay';
+
+      Object.keys(jsonsUrl).forEach(function(key) {
+
+        if (key == 'sessions') {
```
aayusharora: use `===`

src/backend/dist.js
```diff
@@ -522,15 +547,44 @@ module.exports = {
...
+          jsonsUrl[key] = jsonsUrl[key] + '?include=track,microlocation,session-type,speakers&fields[track]=id,name&fields[speaker]=id,name&fields[microlocation]=id,name&page[size]=0';
```
aayusharora: break this code, it is looking clumsy

src/backend/dist.js
```diff
@@ -522,15 +547,44 @@ module.exports = {
...
+        }
+
+        else if (key == 'tracks') {
```
aayusharora: === here

src/backend/dist.js
```diff
@@ -522,15 +547,44 @@ module.exports = {
...
+      'tracks': '',
+      'microlocations': '',
+      'event': '',
+    };
+
+    Object.keys(jsonsUrl).forEach(function(key) {
+      jsonsUrl[key] = endpoint + '/' + key;
+    });
+
+    if(endpoint.search('eventyay') != -1) {
```
aayusharora: can you explain the logic here ?

src/backend/dist.js
Princu7: Nothing much. Just thought that this way a decent amount of time for the request module to wait for a response from a link before timing out and throwing an error that the server is not responding and the user is most probably offline.

src/backend/dist.js
Princu7: Thanks for pointing out!

src/backend/dist.js
Princu7: :+1:

src/backend/dist.js
Princu7: :+1:

src/backend/dist.js
Princu7: Sure! The user can generate the event website using the API endpoint. Now, this API endpoint could be either the raw GitHub link or the new eventyay server link. The raw Github links directly return the whole content of the JSON file when accessed.

But, when the user has provided a link on the eventyay server, then the response is no longer the same because of the migration to the new JSON API model. We have to add certain query parameters to the request URL and then parse it using some software to get the same content back.

Like accessing sessions data from the GitHub URL is simple and can be simply fetched like
https://raw.githubusercontent.com/fossasia/open-event/master/sample/FOSSASIA16/sessions
But, in case the user has provided the server endpoint link, then to get the whole data as above, we have to send the query in a modified manner and then deserialize the JSON API response received
Request URL:
http://api.eventyay.com/v1/events/69/sessions?include=track,microlocation,session-type,speakers&fields[track]=id,name&fields[speaker]=id,name&fields[microlocation]=id,name&page[size]=0
After we receive a response from this URL, we have to deserialize the JSON API response back to get the desired data.

So this code checks whether the API endpoint provided by the user is the raw GitHub link or the eventyay one and accordingly construct the request URL which we have to use to get the desired content and then handle the response accordingly! Thanks!

Princu7:
@aayusharora Made the requested changes! Please review again. Thanks :)

mariobehling:
Merging as Work in Progress.

PR #1453, opened 2017-08-21
Add Selenium tests for checking the font color of session elements on…

### LINKED ISSUE #1311: Test cases for Selenium testing of the website

Princu7:
… different pages

Fixes #1311, #1217 completely

Changes:
* Adds selenium tests for checking the font color of session elements on different pages (tracks, rooms and schedule pages). Takes two sessions. One of their font-color should be white and the other font-color should be black. Checks the color equality on different pages.

### DIFF
--- a/src/selenium/basePage.js
+++ b/src/selenium/basePage.js
@@ -65,6 +65,16 @@ var BasePage = {
     return Promise.all(promiseArr);
   },
 
+  getSessionElemsColor: function() {
+    var self = this;
+    var sessionElemIdArr = ['title-3014', 'title-2941'];
+    var colorPromArr = sessionElemIdArr.map(function(sessionElemId) {
+      return self.find(By.id(sessionElemId)).getCssValue('color');
+    });
+
+    return Promise.all(colorPromArr);
+  },
+
   checkDownButton: function() {
     var self = this;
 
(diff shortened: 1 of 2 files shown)


aayusharora:
Tests seems ok @Princu7

PR #1460, opened 2017-08-25
Enabled web app generation without any need to refresh the page

### LINKED ISSUE #1449: Application gets stuck when generate webapp button is clicked second time without refreshing the page.
**Actual Behaviour**

The application gets stuck when generate webapp button is clicked second time after completion of first process either with wrong API endpoint link or with right one.

**Expected Behaviour**

Generate webapp button should be permanently disabled after generation of website , so that page needs to be refreshed after one use.
Or proper website and errors should be thrown second time also depending on request parameters.

**Steps to reproduce it**

Generate a webapp with proper endpoint link , after its creation press generate webapp button again with same API end point link or some other.

**Screenshots of the issue**

Where-ever possible add a screenshot of the issue.
screenshot from 2017-08-19 19-35-00

Yes,I would like to work on it.

agbilotia1998:
Fixes #1449 

Changes: Removed disabled attribute once generation of webapp is complete.

Link to review:
https://openeveweb.herokuapp.com/


agbilotia1998:
@princu7 please review.

Princu7:
@agbilotia1998 Thanks for the PR :)
Can you please elaborate more on what the PR does? I am getting a bit confused :sweat_smile: 
I generated an event. After generating it, I can still not change any data source, session style or the email address in the form. They are in the disabled form. Please check. Thanks!

agbilotia1998:
@Princu7 , some other branch of my repository was deployed to heroku app for other PR testing.I have deployed the current branch , please review :).

Princu7:
@agbilotia1998 Oh. No problem. It looks nice. Now, we can generate an event via the API endpoint and then generate an event via the zip upload and it works without refreshing the page! But I think that there is a slight issue that we can't generate two events using the API endpoint in succession. It gets stuck at some point. 

Just update this PR by merging in the latest changes. And modify the PR description to "Partially fixes the issue #number". We can solve the other issue which I mentioned in the another PR. Thanks!

mariobehling:
Thank you!

Issue #1456, opened 2017-08-24
Add more samples to the live site


Princu7:
As discussed on the gitter chat, it would be cool to have more samples for display on the live site. These are the samples which are missing:

* Google IO 2017
* PyCon 2017
* FOSSASIA 2015
* Mozilla All Hands 2016
* RedHat 2017
* NextCloud Conf 2017 [Urgent]

Some of the samples have some problems due to which they are unable to be generated right now. I will try to look into it. Thanks!

PR #1462, opened 2017-08-25
Add Nextcloud event to the live site manually

Princu7:
Partially fixes the issue #1456 

**Changes**:
* Adds Nextcloud event to the live site manually. It is added manually instead of auto deploying since the samples are in an inconsistent state now and the auto deployment of the events on each PR merge is not working. A couple of samples are in new attributes while the rest of them are in old attributes. We would have to update them all.

**Test Website**
http://princu7.github.io/open-event-webapp

@mariobehling @niranjan94 @aayusharora  Please review. Thanks :)


mariobehling:
Why manually?

Please automatically generate it using the API endpoint: https://eventyay.com/api/v1/events/181

Princu7:
@mariobehling Yes, we are using the same endpoint but adding the event site manually instead of automatically on each PR. It is being done since we are in the process of migrating from the old attributes to the new attributes. Since the nextcloud event was created on the old eventyay server, it uses the old attributes. Once all the samples have been updated successfully, we can start auto deploying the events again.
I added the nextcloud event manually since it was an urgent issue. Please guide. Thanks! @mariobehling @niranjan94

mariobehling:
What about the FOSSASIA 2017 sample then? We are not using the sample and also the eventyay deployment?

PR #1466, opened 2017-09-04
Added PyCon17 to the live site

### LINKED ISSUE #1456: Add more samples to the live site

Princu7:
Fixes #1456 

Changes:
* Adds PyCon 17 to the live site. The added site contains single session pages since there was only one event having single page session. So, in order to bring more equality, adding the single page session version of this event.


Princu7:
**Test Website**
https://princu7.github.io/open-event-webapp/

@aayusharora  @mariobehling @geekyd @sumedh123 Please review. Thanks :)

geekyd:
@Princu7 There is one thing that I want to point out, when the user clicks on the a session from speakers page shouldn't the individual session should open instead of taking the user to tracks page in case of single session page events?

Princu7:
@geekyd Yeah. That would be more apt. Will open a new issue for it and try to solve it. Thanks!

PR #1479, opened 2017-09-29
Removed preview and download button on source change

### LINKED ISSUE #1036: Changing the data source is not disabling the Generate WebApp Button and is not hiding preview and download button
@aayusharora 
Suppose we have ,
upload

Now everything works fine until here..
next if you change your data source ,
endpoint
Here even if the endpoint is not given , 
1. the generate webapp button is not disabled
2. both the preview and download buttons are not hidden.
if you click on preview it loads the data from the JSON zip file that was uploaded and generates the site , which actually should not happen since your current data source is api endpoint!

I would like to work on this issue!

agbilotia1998:
Fixes #1036 

Screenshots for the change: 
screenshot from 2017-09-29 19-24-08

### DIFF
--- a/src/www/js/form.js
+++ b/src/www/js/form.js
@@ -154,6 +154,9 @@ $(document).ready(function () {
         if ($(this).val() === 'jsonupload') {
           $('#jsonupload-input').show(100);
           $('#eventapi-input').hide(100);
+          $('#btnLive').hide();
+          $('#btnDownload').hide();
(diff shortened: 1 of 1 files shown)


geekyd:
@agbilotia1998 Please check why travis is failing.

agbilotia1998:
Please rebuild travis , I think the issue was with travis itself as it stopped during npm install only.

PR #1502, opened 2017-10-16
Added error catch statement

### LINKED ISSUE #1501: Missing error catch statement in serverTest.js
**Actual Behaviour**

No error log is returned if 'Checking the title of the page' test fails.

**Expected Behaviour**

It should catch the error. 
**Steps to reproduce it**
Change the API such that fossasia webapp has title null and generate webapp with that api changing the endpoint in serverTest.js

Yes.

agbilotia1998:
Fixes #1501 
Changes: Added catch statement for error in title test.

### DIFF
--- a/test/serverTest.js
+++ b/test/serverTest.js
@@ -409,7 +409,7 @@ describe("Running Selenium tests on Chrome Driver", function() {
       }).catch(function(err) {
         done(err);
       });
-    })
+    });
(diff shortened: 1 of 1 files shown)


agbilotia1998:
@geekyd please review.

mariobehling:
Please sync first.

PR #1508, opened 2017-10-19
Proxy issue while installing dependencies.

agbilotia1998:
Added note in docs for avoiding local machine installation behind a proxy network.


agbilotia1998:
@geekyd please review.

niranjan94:
@agbilotia1998 this is not right. Its upto the user to configure proxy correctly for npm/bower. So, this line isn't needed. Closing PR.

agbilotia1998:
@niranjan94 in npm even if you set the proxy correctly node gyp rebuild fails due to proxy configuration.

PR #1600, opened 2018-01-07
Removed errors while speakers image modulation using sharp

### LINKED ISSUE #1599: Console errors while speaker's image resizing using sharp
**Actual Behaviour**

Error is thrown while speaker's image conversion.

**Expected Behaviour**

No error should appear in console.

**Steps to reproduce it**

Generate any sample locally and see the console statements.

**Log for the issue**
`Error: Input file is missing or of an unsupported image format`

agbilotia1998:
Fixes #1599 

Changes: `fs.readdir` was also reading thumbnails folder as the image to be resized.Removing thumbnails directory from the modulation list removed the console errors as it is a directory not an image.

### DIFF
--- a/src/backend/dist.js
+++ b/src/backend/dist.js
@@ -241,7 +241,8 @@ var resizeSponsors = function(dir, socket, done) {
 };
 
 var resizeSpeakers = function(dir, socket, done) {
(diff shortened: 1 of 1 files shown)


agbilotia1998:
@geekyd please review!

PR #1606, opened 2018-01-08
Fixes #1591: Fix hover issue of overlay

### LINKED ISSUE #1591: Generator page not visible when hovering over email choices on page.
**Actual Behaviour**

Generator page not visible when hovering over email choices on page.

**Expected Behaviour**

Generator page should be visible on hovering.

abishekvashok:
Fixes hover display of overlay. Now it wont display when we hover over
the suggestions offered by web browsers. Added a 3 second timeout to the
hover display so that we have sufficient time to pick the suggestion
before it reappears

Fixes #1591 

live link to preview: https://powerful-temple-94510.herokuapp.com/

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -153,10 +153,6 @@
      background-image: url(chill.jpg);
     }
     
-    body:hover .overlay{
-    opacity: 1;
-    }
-    
     @media(max-width: 768px){
       .custom-navtitle, .custom-menubutton {
           font-size: 15px;
@@ -315,5 +311,20 @@
     <span> <a id="deploy-link"> </a> </span>
   </div>
  </div>
+  <script>
+    var browserSuggestionTweak;
+    $(document).on('mouseover', function () {
+      /* Clearing any timeout set under browserSuggestionTweak */
+      clearTimeout(browserSuggestionTweak);
+      $('.overlay')[0].style.opacity = 1;
+    });
+    $(document).on('mouseout', function () {
+      /* Set a timeout as immediatly displaying the overlay when suggestions
+      appear in the browser will be very bad.*/
+      browserSuggestionTweak = setTimeout(function () {
+        $('.overlay')[0].style.opacity = "";
+      }, 3000);
+    });
+  </script>
 </body>
 </html>


abishekvashok:
@fossasia/open-event-web-app-admin @geekyd @fossasia/fossasia-gci  pls review

src/www/index.html
```diff
@@ -178,6 +174,21 @@
       }
     }
   </style>
+  <script>
```
geekyd: Add this at the end inside body.

src/www/index.html
```diff
@@ -178,6 +174,21 @@
...
+        /* Clearing any timeout set under browserSuggestionTweak */
+        clearTimeout(browserSuggestionTweak);
+        $('.overlay')[0].style.opacity = 1;
+    });
+    $(document).on('mouseout', function () {
+        /* Set a timeout as immediatly displaying the overlay when suggestions
+        appear in the browser will be very bad.*/
+        browserSuggestionTweak = setTimeout(function () {
+            $('.overlay')[0].style.opacity = "";
+        }, 3000);
```
geekyd: Lets not add timeout. Let it be instantaneous.

abishekvashok: Making it instantaneous proovides this bug @geekyd. If you could pls see the issue #1591

src/www/index.html
abishekvashok: sure

abishekvashok:
@geekyd done, can you recheck now pls?

src/www/index.html
```diff
@@ -315,5 +311,20 @@
     <span> <a id="deploy-link"> </a> </span>
   </div>
  </div>
+  <script>
```
geekyd: @abishekvashok Check this 
```
    $(document).on('mouseover', function () {
        /* Clearing any timeout set under browserSuggestionTweak */
        $('.overlay')[0].style.opacity = 1;
    });
    $(document).on('mouseout', function () {
        /* Set a timeout as immediatly displaying the overlay when suggestions
        appear in the browser will be very bad.*/
        $('.overlay')[0].style.opacity = "";
    })
```

abishekvashok: It will still produce the bug @geekyd 
I tried that

geekyd: @abishekvashok This problem exists for firefox. In chrome It flickers instead of the opacity=1;
If the user is selecting the emails for more than 3 seconds then the problem will be same. What do you think about it? How should we take care of that?

geekyd:
Closing this due to issue https://github.com/fossasia/open-event-webapp/issues/1608, please take up some other issue. Thanks :)

Issue #1631, opened 2018-01-09
Unknown bug causes menu bar to disappear


0x48piraj:
**Actual Behaviour**

**Unknown Bug** causes menu bar to disappear.

**Expected Behaviour**

Should be bug-free.

**Steps to reproduce it**

![openweb-app-gen-bug](https://user-images.githubusercontent.com/5800726/34729028-95aa2846-f581-11e7-8b7c-323b6aad1cec.gif)

**Log for the issue**

```
<div class="custom-menu-cont">
    <div class="custom-menu">
      <div class="hide-arrow"></div>
      <div class="custom-menu-cont">
        <div class="custom-menu-item">
          <a class="custom-icon" href="http://github.com/fossasia/open-event-webapp" target="_blank"><img src="./icons/code.png">
          <p class="custom-title">Code</p></a>
        </div>
        <div class="custom-menu-item">
          <a class="custom-icon" href="http://github.com/fossasia/open-event-webapp/issues" target="_blank"><img src="./icons/bug.png">
          <p class="custom-title">Bug Report</p></a>
        </div>
        <div class="custom-menu-item">
          <a class="custom-icon" href="http://blog.fossasia.org" target="_blank"><img src="./icons/blog.png">
          <p class="custom-title">Blogs</p></a>
        </div>
        <hr style="margin-bottom: 0; margin-top: 10px;">
        <div class="custom-menu-item">
          <a class="custom-icon" href="https://eventyay.com/" target="_blank"><img src="./icons/eventyay.png">
          <p class="custom-title">Eventyay</p></a>
        </div>
        <div class="custom-menu-item">
          <a class="custom-icon" href="https://susper.com/" target="_blank"><img src="./icons/susper.png">
          <p class="custom-title">Susper</p></a>
        </div>
        <div class="custom-menu-item">
          <a class="custom-icon" href="https://chat.susi.ai/" target="_blank"><img src="./icons/susi.png">
          <p class="custom-title">Susi</p></a>
        </div>
        <div class="custom-menu-item">
          <a class="custom-icon" href="https://fossasia.org/" target="_blank"><img src="./icons/fossasia.png">
          <p class="custom-title">FOSSASIA</p></a>
        </div>
        <div class="custom-menu-item">
          <a class="custom-icon" href="https://loklak.org/" target="_blank"><img src="./icons/loklak.png">
          <p class="custom-title">loklak</p></a>
        </div>
      </div>
    </div>
[...]

PR #1636, opened 2018-01-10
Fixed Issue (#1631). added hidden class to the div which earlier just had custom-menu-content class

tushargr:
… had custom-menu-content class

Fixes #[1631. Note: This will automatically close the issue. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Changes: [Added "hidden" class to the div which earlier just had one class- "custom-menu-content" in index.html    deployment link- https://warm-headland-17924.herokuapp.com/]

Screenshots for the change:

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -168,7 +168,7 @@
   <div class="custom-menu-cont hidden">
     <div class="custom-menu">
       <div class="hide-arrow"></div>
-      <div class="custom-menu-cont">
+      <div class="custom-menu-cont hidden">
         <div class="custom-menu-item">
           <a class="custom-icon" href="http://github.com/fossasia/open-event-webapp" target="_blank"><img src="./icons/code.png">
           <p class="custom-title">Code</p></a>


tushargr:
My deployment link- https://warm-headland-17924.herokuapp.com/
The menu bar button is now working correctly.

src/www/index.html
```diff
@@ -168,7 +168,7 @@
   <div class="custom-menu-cont hidden">
     <div class="custom-menu">
       <div class="hide-arrow"></div>
-      <div class="custom-menu-cont">
+      <div class="custom-menu-cont hidden">
```
geekyd: Remove `custom-menu-cont` from here.

tushargr:
@geekyd if i remove the hidden the toggle in form.js won't work. Actually initially both are hidden, when someone clicks the button, both div get toggled and thus become visible. Isn't it the required behavior

src/www/index.html
geekyd: @tushargr Try removing both `hidden` & `custom-menu-cont` class

PR #1672, opened 2018-01-18
Removed horizontal scroll in schedule page when the number of columns are less than or equal to 3.

### LINKED ISSUE #1472: Schedule: Hide horizontal scroll when there are 3 or less columns
**Actual Behaviour**
screen shot 2017-09-28 at 5 12 59 pm

**Expected Behaviour**
There should be no horizontal scroll when the number of columns are 3 or less

shubham3211:
Fixes #1472
open-event-webapp : https://intense-retreat-31089.herokuapp.com/


geekyd:
@shubham3211 We are setting the width in the backend. Please check fold.js https://github.com/fossasia/open-event-webapp/blob/development/src/backend/fold.js#L700

PR #1772, opened 2018-03-13
Google fonts removed

### LINKED ISSUE #1751: Make fonts work on the deployment itself - avoid getting fonts from external services
We are using Google Fonts. Instead of getting fonts from the Google service implement the fonts statically as part of the web app.

shubham3211:
Webapp-generator:-https://tranquil-bayou-91099.herokuapp.com/

Changes:

1.)Makes the Google font Open Sans work offline now.
2.)Made appropriate changes to make sure all browsers are supported.
Did the required settings with the help of this site :- https://google-webfonts-helper.herokuapp.com/fonts/open-sans   . Tried many different variations and this one was the closest to the current one used on the site. 

Fixes #1751


mariobehling:
Thanks for the PR, but how can we get closer to the previous look? Testing it more thouroughly now. the result is not as desired yet. We need the previous font to work as part of the site.

It actually looks very different now: https://2018.fossasia.org/event/schedule.html

mariobehling:
I have to revert this PR to keep it more easily readable. Could you please continue to help with this issue and provide a test page as you progress? Thank you for your contributions!

PR #1893, opened 2018-05-26
Improves UI for filters

shubham3211:
Hover over the filter to see the changes

Sample:-https://shubham3211.github.io/facebook6/
Fixes ##1890


geekyd:
@shubham3211 Please use proper commit messages, It really helps. I would recommend you to go though https://chris.beams.io/posts/git-commit/. Thanks

src/backend/_scss/application.scss
```diff
@@ -848,10 +849,30 @@ a {
...
 .room-info {
   margin-bottom: 5px;
+  &:hover {
+    .room-name {
+      color: $light-blue;
+    }
+    .titlecolor {
+      height: 14px;
+      margin-right: 0;
+      width: 14px;
```
joeytwiddle: An alternative to setting specific widths and heights could be:
```
      transform: scale(1.16);
```
That should cause the circle to expand about its centre.

`transform:scale()` does not affect the width+height of the element (or its container) in the page, only how it is displayed.

joeytwiddle:
A good commit message could be:
```
git commit -a -m "Highlight track and room links when hovered"
```

src/backend/_scss/application.scss
```diff
@@ -848,10 +849,30 @@ a {
...
+    }
+  }
```
joeytwiddle: I have a concern: The proposed highlight activates when the pointer is hovered _anywhere_ on the `<div>`. But clicking on the `<div>` does not always trigger an action. This will be confusing for users.

To get action, you have to click on the text: `<span class=".track-name">`. (The mouse reconfirms this: when you hover over the text, the pointer changes to a hand. I was testing this on the Tracks page.)

A simple solution is:
1. Move the highlight so it only appears when the mouse is hovered over the text:
```scss
    .track-name:hover {
      color: $light-blue;
    }
    .room-name:hover {
      color: $light-blue;
    }
```

But unfortunately this will not make the title circle grow when the text is hovered. (We [cannot use](https://stackoverflow.com/questions/1817792/is-there-a-previous-sibling-css-selector#comment84623822_15503444) a previous-sibling selector.)

If you want the title circle to grow, there are alternatives. But they change the HTML or Javascript, so they might require some extra work / testing:

2. Wrap the two `<span>`s in a container `<span>`, then apply hover styling to that (similar to what you have now, but `:hover`).
3. or, keep the styling you have now, but change the Javascript so that the filter will activate if the user clicks anywhere on the container `<div>`.

Sorry!

Issue #1881, opened 2018-05-24
Minimize duplication of code in templates


geekyd:
**Actual Behaviour**
Currently the templates are using the same logic for search, filter etc, all the templates have duplicate code. 

**Expected Behaviour**
We should add a separate JS file to reduce the code duplication in the template files

agbilotia1998:
I am working on this.

shubham3211:
@agbilotia1998 are you working on this.

agbilotia1998:
Ummm, you can take this up, right now I am working on a different issue.

shubham3211:
ok :+1:

PR #1894, opened 2018-05-27
Adds video support to v1 web apps

### LINKED ISSUE #1892: Add video stream feature to v1 webapps
**Actual Behaviour**

Currently, only v2 apps support video stream of the sessions.

**Expected Behaviour**

The same implementation should be applied to v1 also.

Yes.

agbilotia1998:
#### Checklist

Adds video stream feature for sessions in v1 web apps

Fixes #1892 
Preview Link: https://agbilotia1998.github.io/Fossasia2018

### DIFF
--- a/src/backend/templates/rooms.hbs
+++ b/src/backend/templates/rooms.hbs
@@ -306,6 +306,7 @@
     <script src="./js/jquery.min.js"></script>
     <script src="./js/bootstrap.min.js" type="text/javascript"></script>
     <script src="./js/rooms.min.js" type="text/javascript"></script>
+    <script src="./js/main.min.js" type="text/javascript"></script>
     <script type="text/javascript">
       var eventName = "{{eventurls.name}}";
       var flag = 0;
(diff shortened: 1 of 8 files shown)


src/backend/templates/rooms.hbs
```diff
@@ -851,6 +851,19 @@
...
           let isVideoDisplayed = $('#desc-' + div).hasClass('in');
+
+          if (videoURL.indexOf('v=') !== -1 && videoURL.indexOf('&feature=youtu.be') === -1) {
+            videoURL = videoURL.split('v=')[1];
+          }
+          else if (videoURL.indexOf('v=') !== -1 && videoURL.indexOf('&feature=youtu.be') !== -1) {
+            const endIndex = videoURL.indexOf('&feature=youtu.be');
+            const startIndex = videoURL.indexOf('v=');
+            videoURL = videoURL.substring(startIndex+2, endIndex);
+          }
```
joeytwiddle: You could save a few lines, if you were willing to use a regexp to strip any unwanted param or hash.
```js
if (videoUrl.indexOf('v=') !== -1) {
  videoUrl = videoUrl.split('v=')[1].replace(/[&#].*/, '');
}
```
Pros and cons! :wink:

(I was also wondering if `new URL(videoUrl).searchParams.get('v')` might work, but that's [not yet](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) available in IE or Edge.)

src/backend/templates/tracks.hbs
```diff
@@ -837,6 +837,19 @@
...
+            videoURL = videoURL.split('v=')[1];
+          }
+          else if (videoURL.indexOf('v=') !== -1 && videoURL.indexOf('&feature=youtu.be') !== -1) {
+            const endIndex = videoURL.indexOf('&feature=youtu.be');
+            const startIndex = videoURL.indexOf('v=');
+            videoURL = videoURL.substring(startIndex+2, endIndex);
+          }
+          else if (videoURL.indexOf('https://youtu.be/') !== -1) {
+            videoURL = videoURL.split('https://youtu.be/')[1];
+          }
```
joeytwiddle: Could you put this code in a common library script, so that it won't be [repeated 3 times](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))?

It's often easier to do that when writing the code than it is to do it [later](https://github.com/fossasia/open-event-webapp/issues/1881#issuecomment-391686292). (In the future, we may not be so sure that the code is an exact match.)

src/backend/templates/rooms.hbs
agbilotia1998: Thanks for the `regex` solution, I will make the changes. :smile:

agbilotia1998:
Too many comments by hound and codacy, closing this and opening a new PR.

PR #1905, opened 2018-06-01
Implements task queue in the project

### LINKED ISSUE #1856: Implement task queue in the project.
**Actual Behaviour**

In the current implementation of open-event-web app multiple users cannot generate an event at the same time, only the recent triggered event is generated and generation of all the former ones is halted.

**Expected Behaviour**

We need to enable multiple web-apps generations by implementing queue data structure for the series of requests triggered by users from the open-event platform. This would prevent the loss of requests from the user and enable the creation of event for every request.

**Steps to reproduce it**

Generate any two web apps simultaneously.

Yes.

agbilotia1998:
#### Checklist

Implements task queue in the project, which enables multiple web app generation without the loss of request from the user.

- Implemented queue backed by Redis server.

Fixes #1856 
Link to test the changes: https://openeveweb.herokuapp.com/

API endpoints for testing:
 - v1:
`https://eventyay.com/api/v1/events/200`

- v2:
`https://raw.githubusercontent.com/fossasia/open-event/master/sample/MozillaAllHands17`

-PS: Heroku Redis server addon needs to be enabled before merging this PR.

### DIFF
--- a/src/backend/generator.js
+++ b/src/backend/generator.js
@@ -248,7 +248,9 @@ exports.createDistDir = function(req, socket, callback) {
       fs.remove(distHelper.distPath + '/' + appFolder, (err) => {
         if (err !== null) {
           logger.addLog('Error', 'Failed to clean up the previously existing temporary folders', socket, err);
+          callback(null);
           console.log(err);
+          done(err);
         }
         logger.addLog('Success', 'Successfully cleaned up the temporary folders', socket);
         done(null, 'clean');
@@ -274,6 +276,7 @@ exports.createDistDir = function(req, socket, callback) {
         if (copyerr !== null) {
           console.log(copyerr);
           logger.addLog('Error', 'Error occured while copying assets into the appFolder', socket, copyerr);
+          callback(null);
           return socket.emit('live.error', {donePercent: 30, status: 'Error in Copying assets'});
         }
         logger.addLog('Success', 'Assets were successfully copied', socket);
@@ -289,6 +292,7 @@ exports.createDistDir = function(req, socket, callback) {
         console.log('============================Moving contents from dependency folder and deleting the dependency folder');
         if (copyerr !== null) {
           logger.addLog('Error', 'Error while reading directory', socket, copyerr);
+          callback(null);
           console.log(copyerr);
           return socket.emit('live.error', {donePercent: 45, status: 'Error in moving files from dependency folder'});
         }
@@ -310,6 +314,7 @@ exports.createDistDir = function(req, socket, callback) {
           distHelper.copyUploads(appFolder, socket, uploadsId, function(err) {
             if (err) {
               console.log(err);
+              callback(null);
               done(err);
             } else {
               done(null, 'copyUploads');
@@ -322,6 +327,7 @@ exports.createDistDir = function(req, socket, callback) {
           distHelper.fetchApiJsons(appFolder, req.body.apiendpoint, socket, (err) => {
             if (err !== null) {
               console.log(err);
+              callback(null);
               return done(err);
             }
             logger.addLog('Success', 'All jsons have been successfully downloaded', socket);
@@ -363,6 +369,7 @@ exports.createDistDir = function(req, socket, callback) {
           if (emit) {
             socket.emit('live.error', {status: 'Error in Compiling SASS'});
           }
+          callback(null);
         }
       });
     },
@@ -383,6 +390,7 @@ exports.createDistDir = function(req, socket, callback) {
           if (emit) {
             socket.emit('live.error', {status: 'Error in read contents of zip'});
           }
+          callback(null);
           return done(error);
         }
 
@@ -402,6 +410,7 @@ exports.createDistDir = function(req, socket, callback) {
             distHelper.optimizeLogo(logoPath, socket, function(err, pad) {
               if (err) {
                 console.log(err);
+                callback(null);
                 return done(err);
               }
               jsonData.navpad = pad;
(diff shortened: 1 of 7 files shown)


agbilotia1998:
@geekyd please review.

geekyd:
@agbilotia1998 Please check the codacy issues

agbilotia1998:
@geekyd can you please mention the test data for which event generation failed, as I tried to handle all the errors and tested multiple cases for `Error in read contents of zip` and the events are getting generated after this.

agbilotia1998:
Yes sure, I will add the documentation for the changes.

geekyd:
@agbilotia1998 I used the API you provided in the description.

agbilotia1998:
@geekyd I tested for several kinds of error and it worked fine. Also, I checked Heroku logs but couldn't find anything there as well, it would be really helpful if you mention the error encountered. Thanks :smile:

agbilotia1998:
@geekyd can you please help me here, I tested for null zip upload as well but the app was working fine. Maybe when you tested it resulted into memory exceed for Heroku, actually this endpoint `https://eventyay.com/api/v1/events/275` is for FOSSASIA Summit 2018 which is a large event, so on Heroku account with no dynos enabled, multiple times generating this app may result into the memory related error.

agbilotia1998:
@geekyd can you please help me here, I tested for null zip upload as well but the app was working fine. Maybe when you tested it resulted into memory exceed for Heroku, actually this endpoint https://eventyay.com/api/v1/events/275 is for FOSSASIA Summit 2018 which is a large event, so on Heroku account with no dynos enabled, multiple times generating this app may result into the memory related error.
@ParthS007 @pradeepgangwar @srv-twry can also checkout if I am missing something?

pradeepgangwar:
@agbilotia1998 It works for me. :+1:

agbilotia1998:
@geekyd @ParthS007 @mayank8318 I have resolved the conflicts, please check.
Link to test the changes: https://openeveweb.herokuapp.com/

src/backend/generator.js
```diff
@@ -248,6 +248,7 @@ exports.createDistDir = function(req, socket, callback) {
       fs.remove(distHelper.distPath + '/' + appFolder, (err) => {
         if (err !== null) {
           logger.addLog('Error', 'Failed to clean up the previously existing temporary folders', socket, err);
+          callback(null);
```
joeytwiddle: If you don't want the process to continue, you could `return` here.

src/backend/generator.js
```diff
@@ -542,13 +553,15 @@ exports.createDistDir = function(req, socket, callback) {
...
           return done(err);
         }
         distHelper.copyServiceWorker(appFolder, hashObj.hash, function(error) {
           if (error) {
             console.log(error);
             logger.addLog('Error', 'Error occured while copying service worker file', socket, err);
-            return done(error);
+            callback(null);
+            return done(err);
```
joeytwiddle: I think this should be `return done(error);`

And the `addLog()` line above should use `error` too!

(By this point in the code, `err` is unset.)

src/backend/generator.js
```diff
@@ -565,6 +578,7 @@ exports.createDistDir = function(req, socket, callback) {
         if (err) {
           console.log(err);
           logger.addLog('Error', 'Error occured while copying manifest file', socket, err);
+          callback(null);
```
joeytwiddle: Usually `return done(err);` would be enough here, but only if this `async.series()` had a [final callback function]((https://caolan.github.io/async/docs.html#series)). At present it does not.

But perhaps fixing that might be preferable to adding a lot of `callback(null)`s!

src/backend/generator.js
joeytwiddle: It looks like there is a missing `return` here. (Not from this PR, but from the original code.)

At present it will log an error "Failed to clean up" but then log "Successfully cleaned up" and continue with the process.

src/backend/generator.js
agbilotia1998: Yes, that should be `error` thanks.

src/backend/generator.js
agbilotia1998: Yes, async.series() is having a final callback, but it should be called at the time of the error too. callback(null) is not for the async series in the generator but is for the task-queue.

src/backend/generator.js
agbilotia1998: Yes, `return done(err)` should be added here.

src/backend/generator.js
joeytwiddle: That's right. But instead of calling it for every error, we can just add this to the end of `async.series`:

```js
  ], (err, result) => {
    if (err) {
      // One of the handlers above has called done(err)
      // Usually we would tell the parent that we had a problem:
      //return callback(err);
      // But at present, we just tell them that we are finished:
      return callback(null);
    }
    // Success!
    // Usually we would tell the parent that we are finished
    //callback(null, result);
    // But at present, this is being done in the last series function above. So we do nothing here.
  });
```

This means we don't have to call `callback(null)` from the series functions, but we _do_ have to call `return done(err)` from them.

joeytwiddle: In fact, usually `async.series()` will simply end with:
```js
  ], callback);
```
but we cannot do that here, because our flow is a bit different. (We don't pass the error back, and the result has already been passed back.)

agbilotia1998: Yes, I agree with you. This callback would be called after all the functions are completed, right? And the `callback(null)` would be written once only.

src/backend/generator.js
agbilotia1998: Added, thanks :)

src/backend/generator.js
joeytwiddle: Correct.

Thet final function (outside the array) gets called:
- immediately if an error occurred
- or at the very end, if all the steps succeeded.

Issue #1882, opened 2018-05-24
Increase speed of development iteration


joeytwiddle:
When a developer writes new code, it is helpful if they can test that code really quickly.

The faster we can make the code-build-test iteration, the more work each developer will be able to do per hour.

I am not quite sure how to achieve this, so we may want to brainstorm.

1. ~**Improve/automate the build process**: a) Make it so I can build without having to reload the generator page, fill it out, and upload the zip file again. b) Inform me properly when the build ends. Sometimes the progress bar stops at 40% or 70% but when the console shows "SendGridError: Response error" it has actually finished.~

2. **Automatic builds**: We could add a task `npm run watch` which would automatically trigger a new build whenever the source code is changed (e.g. using [nodemon](https://www.npmjs.com/package/nodemon)). This won't be *much* faster, but it will be less effort for the developer. They can read code or make commits instead of manually restarting the build each time they want to test. (If you can create a command that performs a build, I can make a nodemon task for it.)

3. **Incremental builds**: One way to achieve fast rebuilds is with an incremental compiler: when a source file changes, it will build **only** those files in the output which are affected by that source file.

    GNU Make and Node's [Gulp](https://www.npmjs.com/package/gulp) and Create-React-App and Webpack work that way.

    Our build process is quite dynamic, because the event's JSON file is also part of the "source" configuration. But Javascript is a dynamic language, so I think it should be possible to achieve what we want using Gulp.

joeytwiddle:
1. Hey you are right! Problem 1 is not happening for me today. The build is faster and easier. :) (Previously it had been re-downloading and processing the images every time. I don't know what changed, but I do fetch and rebase each day.)

2. Right, we could use `nodemon -i dist -i build ` to ignore changes to those folders, or we could use `nodemon -w src -w test` to watch only those folders.

agbilotia1998:
Fixed with #1931

PR #1931, opened 2018-06-12
Use latest templates

joeytwiddle:
#### Checklist

After editing one of the `.hbs` template files, a developer must restart their server, because the templates are only loaded at boot time.

When I first started working on the project, this was somewhat confusing. I would make changes to the template, and rebuild the site, but my changes did not appear!

This PR loads the templates **from fresh** each time a website is generated.

I imagine this will have minimal effect on the live site, but it will make things easier for developers.

I mentioned this concern in issue #1882.

The PR also deduplicates some of the repeated long lines by extracting that behaviour into a function.

#### Concerns

This approach means we re-register already registered partials each time a site is generated. According to the [handlebars code](https://github.com/wycats/handlebars.js/blob/f21f900cf5eda5782781d7ff0ddeba0cd877a4bc/lib/handlebars/base.js#L56) I don't think this will cause any issues, it will simply overwrite the old partial.

### DIFF
--- a/src/backend/generator.js
+++ b/src/backend/generator.js
@@ -19,6 +19,16 @@ const ftpDeployer = require('./ftpdeploy');
 const app = require('../app');
 let fold;
 
+function registerPartial(name) {
+  const partial = handlebars.compile(fs.readFileSync(__dirname + '/templates/partials/' + name + '.hbs').toString('utf-8'));
+
+  handlebars.registerPartial(name, partial);
+}
+
+function compileTemplate(name) {
+  return handlebars.compile(fs.readFileSync(__dirname + '/templates/' + name + '.hbs').toString('utf-8'));
+}
+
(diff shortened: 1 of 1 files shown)


joeytwiddle:
I tried deploying to Heroku but it produced some NPM errors (which told me to read a logfile which I could not find!)

Codacy has been complaining about issues which _contradict_ the settings in `.eslintrc`

I think I have fixed that by renaming `.eslintrc` to `.eslintrc.json`. Now Codacy will use the project's settings.

joeytwiddle:
If I fix the Codacy issues, then eslint will report errors (and the new code will not be consistent with the existing code in this file).

I am not sure if this project really wants to conform to all of Codacy's requirements.

But if it does, then we should change the `.eslintrc` to match!

PR #1982, opened 2018-06-30
code minimize

### LINKED ISSUE #1881: Minimize duplication of code in templates

shubham3211:
Fixes #1881


shubham3211:
@agbilotia1998 done :+1:

src/backend/assets/js/main.js
```diff
@@ -82,38 +82,80 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
   }
 };
 
-window.handleClientLoad = handleClientLoad;
-window.initClient = initClient;
-window.handleAuthClick = handleAuthClick;
-window.listUpcomingEvents = listUpcomingEvents;
-window.loadVideoAndSlides = loadVideoAndSlides;
+$(document).ready(function () {
+  var filter;
```
agbilotia1998: Replace `var` with `let` or `const` depending on the context.

src/backend/assets/js/main.js
```diff
@@ -82,38 +82,80 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
     const faviconDiv = $('#v' + div);
 
     if (videoURL.indexOf('v=') !== -1) {
-      video = videoURL.split('v=')[1].replace(/[&#].*/, '');
+        video = videoURL.split('v=')[1].replace(/[&#].*/, '');
```
agbilotia1998: The indentation should remain constant in every file of the project. Right now it is `2 spaces`.

Kreijstal: I think we can let a linter handle this one

shubham3211:
@agbilotia1998 made the changes :+1:

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,49 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+        }
+    } else {
+       $('#no-results').remove();
+      }
+  };
+
+  window.mainObj= {
+    noResult:  noResult
+  };
+  window.handleClientLoad = handleClientLoad;
```
agbilotia1998: Can you put all these functions in `mainObj` and make the necessary changes that would make the implementation consistent. Also wouldn't it be good if we rename the object to just `main`?

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,49 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+    filter=eval(filterType);
+    var search=false;
+    $('.fossasia-filter').each(function() {
+      if($(this).is(':visible') && $(this).val().length!==0) {
+        search=true;
+      }
+    });
+    var listFilterLength =$(filter.dateFilter+':visible').length;
+    var calendarFilterLength = $('.room:visible').length;
+    if((calendarMode && calendarFilterLength == false) || (!calendarMode &&listFilterLength == false)) {
```
agbilotia1998: Please give a space after `&&`.

src/backend/assets/js/main.js
shubham3211: @agbilotia1998 I think that there is no need to create a object and i should remove it.

src/backend/assets/js/main.js
```diff
@@ -101,7 +101,7 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
 
     if (!isSlideDisplayed && $('[id="slide-' + div + '"]').length === 0) {
       if (slideURL.indexOf('pdf') !== -1) {
-        speakerDiv.prepend('<iframe id = "slide-' + div + '" class = "iframe col-xs-12 col-sm-12 col-md-12" frameborder="0" src="https://docs.google.com/gview?url=' + slideURL + '&embedded=true"></iframe>');
+       speakerDiv.prepend('<iframe id = "slide-' + div + '" class = "iframe col-xs-12 col-sm-12 col-md-12" frameborder="0" src="https://docs.google.com/gview?url=' + slideURL + '&embedded=true"></iframe>');
```
joeytwiddle: Only 1 space indent here

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,49 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+    dateFilter:'.date-filter'
+  };
+  //store track filters
+  let track ={
+    dateFilter:'.date-filter'
+  };
+  //store schedule filters
+  let schedule={
+    dateFilter:'.day-filter'
+  };
```
joeytwiddle: Shouldn't there be a space before and after every `=`?

I recommend getting an eslint plugin for your editor.

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,49 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+
+  const noResult=function(filterType,calendarMode) {
```
joeytwiddle: Shouldn't there be a space after the `,` comma?

How about the policy of using a verb/action for function names? `showNoResults()`

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,49 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+    });
+    var listFilterLength =$(filter.dateFilter+':visible').length;
+    var calendarFilterLength = $('.room:visible').length;
+    if((calendarMode && calendarFilterLength == false) || (!calendarMode && listFilterLength == false)) {
+      $('#no-results').remove();
+      if(search) {
+        $('.date-list').after("<p style = 'padding-left: 15px; margin-top: 15px; font-weight: bold; color: red' id='no-results'>No matching results found.</p>");
+      } else {
+        $('.date-list').after("<p style = 'padding-left: 15px; margin-top: 15px; font-weight: bold; color: red' id='no-results'>No result found.</p>");
+        }
```
joeytwiddle: If you want to minimize this:
```js
const message = search ? 'No matching results found.' : 'No result found.';
$('.date-list').after("<p style='padding-left: 15px; margin-top: 15px; font-weight: bold; color: red' id='no-results'>" + message + "</p>");
```

(Although I'm not sure if two different messages are really needed anyway. They are almost the same!)

src/backend/assets/js/main.js
joeytwiddle: But something else is wrong here.

The variables `room`, `track` and `schedule` are not used inside this function.

Kreijstal:
@joeytwiddle I think it should be installed on the repo, as well.

src/backend/assets/js/main.js
shubham3211: @joeytwiddle please read schedule.hbs , room.hbs, track.hbs

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,46 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+    dateFilter:'.date-filter'
+  };
+  //store track filters
+  let track = {
+    dateFilter:'.date-filter'
+  };
+  //store schedule filters
+  let schedule = {
+    dateFilter:'.day-filter'
+  };
```
joeytwiddle: OK I see what you are trying to do here. The use of `eval` is strongly unrecommended. Here is what you can do instead:

```js
const allFilters = {
  room: {
    dateFilter: '.date-filter'
  },
  track: {
    dateFilter: '.date-filter'
  },
  schedule: {
    dateFilter: '.date-filter'
  }
};
```

Now we have a well-defined object with the filters in it. And so...

src/backend/assets/js/main.js
```diff
@@ -112,8 +112,46 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+
+  const showNoResult = function(filterType , calendarMode) {
+    filter = eval(filterType);
```
joeytwiddle: Here we can do:

```js
const filter = allFilters[filterType];
```

This avoids using `eval` and is clearer for other developers to understand.

src/backend/templates/schedule.hbs
```diff
@@ -988,7 +988,7 @@
           }
         });
 
-        noResult();
+        main.showNoResult('schedule',isCalendarView);
```
joeytwiddle: Should be:
```js
main.showNoResult('schedule', isCalendarView);
```

PR #1986, opened 2018-07-02
Makes the caption clickable

### LINKED ISSUE #1985: Caption not clickable in overview site.
**Actual Behaviour**

The caption in the overview site is not clickable, unlike the main eventyay site.

**Expected Behaviour**

The caption should also be clickable along with the image.

Yes

agbilotia1998:
#### Checklist

Makes the caption clickable along with the event image similar to eventyay.

Preview link: https://agbilotia1998.github.io/overview-site/
Fixes #1985

### DIFF
--- a/src/backend/overviewSite/index.html
+++ b/src/backend/overviewSite/index.html
@@ -106,8 +106,11 @@
             cursor: pointer;
 
         }
-        .event-holder:hover > .thumbnail > .caption {
+        .event-holder:hover > .thumbnail > a > .caption {
             background-color: #f7f7f7;
+        }
+        .event-holder:hover > .thumbnail > a {
+            text-decoration: none;
         }
         .event-holder hr {
             margin-top: 10px;
@@ -126,7 +129,7 @@
             height: 40px;
             text-overflow: ellipsis;
         }
-        .event-holder .caption {
+        .event-holder > .thumbnail > a > .caption {
             background-color: #FFFFFF;
             transition: background-color 0.5s;
         }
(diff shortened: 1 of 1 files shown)


Kreijstal:
Instead of onclicks methods shouldn't we use event listeners, I believe we're using jQuery already, couldn't we use that instead?

agbilotia1998:
Ummm, we are using jQuery in the overview site just for the bootstrap functionalities, and I think there is no need to separate the script for this small thing. Also, if we add jQuery event listeners I guess we would be tracing the event source for proper redirection but that is also not consistent in our case.

src/backend/overviewSite/index.html
```diff
@@ -174,140 +174,112 @@ <h1>Open Event Web Apps</h1>
...
+                <img src='./fasmall.jpg' onerror="imgError(this, true);">
                 <div class="caption">
                     <h4 class="name">FOSSASIA Summit 2018</h4>
                 </div>
             </div>
         </div>
         <div class="col-md-4 col-xs-12 col-sm-6 event-holder">
-            <div class="thumbnail">
-                <a href='./OpenTechSummit/index.html' target='_blank'>
+            <div class="thumbnail" onclick="window.open('./OpenTechSummit/index.html')">
```
Kreijstal: There's absolutely nothing wrong with the anchor element. There's no need to create an onclick listener to open a new webpage when this can be achieved without javascript, and without popups.

What you should do instead to fix the problem is to to extend the anchor link to include the caption.

```html
 <a href='./OpenTechSummit/index.html' target='_blank'>	
                 <img src='./otssmall.jpg' onerror="imgError(this, true);">	                 <img src='./otssmall.jpg' onerror="imgError(this, true);">
                
                 <div class="caption">	                 <div class="caption">
                    <h4 class="name">Open Tech Summit 2017</h4>	+                    <h4 class="name" onclick="window.location.href=''">Open Tech Summit 2017</h4>
                 </div>
</a>
```

agbilotia1998: I tried doing this, but it disrupted the design. No worries I will find another way to do it.

agbilotia1998:
Made the changes, please review.

PR #1993, opened 2018-07-05
filters code removed from templates

### LINKED ISSUE #1881: Minimize duplication of code in templates

shubham3211:
Fixes #1881


shubham3211:
@agbilotia1998 done :+1:

src/backend/assets/js/main.js
```diff
@@ -112,6 +112,84 @@ const loadVideoAndSlides = function(div, videoURL, slideURL) {
...
+
+  window.main = {
+    trackRoomFilter: trackRoomFilter
+  }
+
+});
+
 window.handleClientLoad = handleClientLoad;
 window.initClient = initClient;
 window.handleAuthClick = handleAuthClick;
```
agbilotia1998: Please move all these functions to `main` object so that the implementation remains consistent.

shubham3211:
@agbilotia1998 done :+1:

agbilotia1998:
@shubham3211 I went through the changes, everything looks good. Please provide the deployment link to the server as we need to test some functionalities manually. Thanks :smile:

shubham3211:
@agbilotia1998 I have updated the sample link , add to calendar is available .

shubham3211:
@agbilotia1998 it is working

agbilotia1998:
Oh may be old page was cached. 
For the `session.hbs` page, we are using the code from `main.js` only. For instance check this - https://github.com/fossasia/open-event-webapp/blob/development/src/backend/templates/session.hbs#L260

shubham3211:
I will do it , I was looking at the old version

shubham3211:
@agbilotia1998 done :+1:

shubham3211:
Thanks 👍

PR #2051, opened 2018-07-28
date filter added

### LINKED ISSUE #2050: Implement date filter on room and track page like schedule page

Yes

shubham3211:
Fixes #2050 
Sample :-  https://shubham3211.github.io/calendar-fix/index.html


shubham3211:
@agbilotia1998 please review.

src/backend/templates/tracks.hbs
```diff
@@ -583,6 +584,16 @@
           }
         });
 
+        $('.date-button').click(function() {
+
```
agbilotia1998: Please remove this extra line.

src/backend/templates/tracks.hbs
```diff
@@ -583,6 +584,16 @@
...
+          let className = $(this).attr('class').split(' ')[0];
+          $(this).addClass('is-active');
```
agbilotia1998: Add a line space here.

src/backend/templates/tracks.hbs
```diff
@@ -583,6 +584,16 @@
...
+          $(this).siblings().removeClass('is-active');
+          main.viewsFilter('track', 'list', className);
+          main.showNoResult('track', false);
+
```
agbilotia1998: Please remove this extra line.

src/backend/assets/js/main.js
```diff
@@ -203,9 +203,40 @@ $(document).ready(function () {
...
+      $(filter.dateFilter).each(function () {
+        if(date !== 'all') {
+          if ($(this).attr('class').split(' ')[0] === date) {
+            $(this).removeClass('hide');
+          } else {
+            $(this).addClass('hide');
+          }
+        } else{
+          $(this).removeClass('hide');
+        }
```
joeytwiddle: I think it might be possible to compress this big `if` down to just two lines, decision logic and action:
```js
const showThing = date === 'all' || $(this).hasClass(date);
$(this).toggleClass('hide', !showThing);
```

(Edit: Oh I see, this is not new code, just moved!)

src/backend/templates/rooms.hbs
```diff
@@ -592,6 +593,16 @@
             $('html, body').animate({scrollTop: 0}, 500);
         });
 
+        $('.date-button').click(function() {
+
+          let className = $(this).attr('class').split(' ')[0];
```
joeytwiddle: Assuming the class will always be at position `0` is a bit fragile!

I think the idiomatic way would be to use a data attribute, instead of a class, when we want to attach simple data to an element.

```js
const className = $(this).attr('data-date');
```

In that case, the element might need to look more like this:
```handlebars
<a href="#" data-date="{{firstSlug}}" class="date-button tabs-nav-link">{{caption}}</a>
```

I might have got that wrong, but I hope you get the idea: `class=` for styling, `data-foo=` for data!

You can add as many `data-` attributes as you like. It is an open namespace for developers to use. (For example, Bootstrap uses things like `data-toggle` to add extra behaviour to an element.)

src/backend/templates/schedule.hbs
```diff
@@ -837,47 +837,16 @@
...
 
         $(this).addClass('is-active');
         $(this).siblings().removeClass('is-active');
 
         if(isCalendarView) {
-          viewsFilter('calendar', className);
+          main.viewsFilter('schedule', 'calendar', className);
         } else {
-          viewsFilter('list', className);
+          main.viewsFilter('schedule', 'list', className);
```
joeytwiddle: Super :)

PR #2083, opened 2018-08-23
Fixes download of large size background image

### LINKED ISSUE #2082: Event picture converted to very large PNG and not showing up on main page
Using the API on https://eventyay.com/api/v1/events/641 does not generate a zip with a proper event picture. Instead there is a very large PNG in the folder that is not showing up.

agbilotia1998:
#### Checklist

Fixes download of large size background image.

Screenshot:
screenshot_2018-08-23 jugaadfest

Test Link: https://open-event-webapp.herokuapp.com/

Fixes #2082

### DIFF
--- a/src/backend/dist.js
+++ b/src/backend/dist.js
@@ -26,6 +26,7 @@ const downloadFile = function(url, file_path, next) {
   });
 
   fileStream.on('finish', function() {
+    next();
     fileStream.close();
   });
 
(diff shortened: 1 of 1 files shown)


src/backend/dist.js
```diff
@@ -26,14 +26,14 @@ const downloadFile = function(url, file_path, next) {
   });
 
   fileStream.on('finish', function() {
+    next();
```
joeytwiddle: You might also want to call `next()` inside the `catch` below, if you want the process to continue after an error, instead of stalling.

Alternatively you could call `next(err)` from the catch, if you want to pass the error back for the caller to handle, in which case you shouldn't really need to log here.

PR #2157, opened 2020-01-05
Fixed the sessions page

Dishebh:
#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://blog.fossasia.org/open-source-developer-guide-and-best-practices-at-fossasia).
- [x] My branch is up-to-date with the Upstream `development` branch.
- [x] I have added necessary documentation (if appropriate)

#### Short description of what this resolves:
Made the sessions page same as the tracks page.

Fixes #2137 

#### Screenshot of the result:
Screenshot from 2020-01-05 17-02-33

Issue #2128, opened 2018-12-09
Fix Travis


mariobehling:
Travis is currently failing. Please help to fix it.

Ananthavijay:
@mariobehling this is not yet fixed. Doing some research. Will open a PR soon
Thanks

Dishebh:
Can you please show the logs for the error? That way, it would be easier to resolve the issue.

kushthedude:
@mariobehling Still not solved

iamareebjamal:
It was disabled instead of being fixed, hence reopening

iamareebjamal:
Cause of test failures 

1. The images in sample are invalid

Dishebh:
> Cause of test failures
> 
> 1. The images in sample are invalid

So, shall I remove the images (which are not being used currently) present in `docs/images`?

iamareebjamal:
No, they need to be fixed

Dishebh:
> No, they need to be fixed

Fixed, how?

iamareebjamal:
I'm talking about image URLs in JSON

Dishebh:
> I'm talking about image URLs in JSON

Can you pls cite the source of json file? Thanks

PR #2121, opened 2018-10-21
date filter added on various pages

shubham3211:
Fixes #2077 
Sample:- https://shubham3211.github.io/google-filter/index.html

PR #2172, opened 2020-02-29
fix: Many recent PRs which broke tests and site

iamareebjamal:
Schedule Calendar switcher broken in #2121 
Track layout and direct link broken in #2157 

For #2128


iamareebjamal:
Finally fixed. Two obscure selenium bugs remain

PR #2305, opened 2020-05-11
Aligned radio buttons

### LINKED ISSUE #2162: Misalignment of radio buttons
**Actual Behaviour**
When clicked on the `GENERATE WEBAPP` button, the radio buttons get misaligned w.r.t. each other.

**Expected Behaviour**
There should be no misalignment between the radio buttons.

**Screenshots of the issue**
Screenshot from 2020-01-06 19-38-21

Yes

jlegha:
#### Checklist

There was a problem of radio buttons which got misaligned when clicked on the GENERATE WEBAPP button. So I fixed this issue and now it works fine.

-The alignment of buttons fixed.

Fixes #2162

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -316,7 +316,7 @@
                               .innerHTML = inpObj.validationMessage; 
                 } else { 
                     document.getElementById("geeks") 
-                              .innerHTML = "Input is ALL RIGHT"; 
+                              .innerHTML = ""; 
                 } 
             } 
           </script>
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -293,16 +293,33 @@
         <!-- changes-2 removed till here -->
         <!-- changes-2 added from here -->
         <ul style="list-style-type:none;postion: relative;">
-          <li>
-            <input name="session" required="" type="radio" value="single" id="single">
-            <label for="single">Single</label>
+            <li>
+            <label for="single" >
+              <input name="session" type="radio" value="single" id="single" class="form-control-static"> 
+              Single
+            </label>
           </li>
 
           <li>
-            <input name="session" type="radio" value="expand" id="expandable">
-            <label for="expandable">Expandable</label>
+            <label for="expandable">
+              <input name="session" type="radio" value="expand" id="expandable" required>
+              Expandable
+            </label>
           </li>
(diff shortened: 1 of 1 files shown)


jlegha:
Can you help me with the issue travis build is failing. I don't what is causing this error.

Dishebh:
> Can you help me with the issue travis build is failing. I don't what is causing this error.



> Can you help me with the issue travis build is failing. I don't what is causing this error.

Did you run all the tests and saw if they were passed? `npm test`

jlegha:
I have ran the test but I don't know how to read log of npm test. I don't what npm test does.

src/www/index.html
```diff
@@ -293,16 +293,33 @@
...
+                var inpObj = document.getElementById("expandable"); 
+                if (!inpObj.checkValidity()) { 
+                    document.getElementById("geeks") 
+                              .innerHTML = inpObj.validationMessage; 
+                } else { 
+                    document.getElementById("geeks") 
+                              .innerHTML = "Input is ALL RIGHT"; 
+                } 
+            } 
+          </script>
```
Dishebh, line 322: What is the need to add an extra script tag here? The validation should work without it, as well. Pls see to it, Refer how the validation works for Email input field, rest all the fields should follow the same steps.

jlegha: If we don't use the script tag here then the red message will be shown right below the radio button again causing the same issue you mentioned above. 
What I am doing here is show the custom error message right after both the radio buttons instead of showing right after the first radio button which will again cause the misalignment of radio buttons.

Dishebh: You have to just show `This field is required` in red at the end of both the radio buttons, as shown by any other fields which are required. If you find an additional script tag used specifically for this purpose, then kindly let me know. I guess, it is better to avoid some extra tags in the index file, when the changes could be implemented without it.

jlegha: "This field is required" won't show below both the fields. It will only show after the first radio button. In the image below you can see that it is showing right after the radio button before the label Single.
image

Dishebh: Yes, and we want this as a result (the one you showed), but without use of script tags. See if it can be implemented, and if not, do let me know.

PR #2369, opened 2020-10-28
fix: Move the video to the top when session is expanded

### LINKED ISSUE #2253: Change location of video to top when session is expanded
The video is currently below the description. It should be the other way around.

Please move the video to the top.

Screenshot from 2020-03-20 10-25-49

djcruz93:
### **Fixes #2253**

    - Earlier when the session was expanded the video rendered below the description
    - This commit moves the video to the top and the other elements render in the same order when session is expanded
    - Move the div to the top which now contains all the elements of the session when expanded

### **Before the changes**
image

### **After the changes**
image

### DIFF
--- a/src/backend/templates/schedule.hbs
+++ b/src/backend/templates/schedule.hbs
@@ -221,26 +221,25 @@
                         </div>
                       </h4>
                       <div id="desc-{{session_id}}" class="collapse">
-                        {{{description}}}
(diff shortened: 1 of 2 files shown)


djcruz93:
@iamareebjamal @mariobehling Can you please review? Let me know if there is any scope of improvement.

iamareebjamal (approved):

PR #2371, opened 2020-10-29
fix: Make width of sessions uniform for schedule and tracks pages

### LINKED ISSUE #2137: Schedule page sessions width and area not same as on tracks.html
Please ensure the sessions page are the same as the tracks.html page, e.g. make it full width using the area available.

Compare schedule page;
screenshot from 2019-02-18 01-58-58

to tracks page. It should look the same!
screenshot from 2019-02-18 01-58-34

djcruz93:
### **Fixes #2137**
- Earlier the session div was inline-block in the schedule page
- Make it block to match with tracks page

### **Before the change**
image

### **After the change**
image

### DIFF
--- a/src/backend/_scss/application.scss
+++ b/src/backend/_scss/application.scss
@@ -1079,7 +1079,6 @@ a {
   }
 
   margin-right: 10px;
-  display: inline-block;
   vertical-align: text-top;
 }
 


djcruz93:
@iamareebjamal can you please review?

iamareebjamal (approved):

PR #2373, opened 2020-10-30
fix: Enable url input field to store the past input data

### LINKED ISSUE #2081: Field is secured to not store data - it should not
The API field is created in a way to not store data. Usually a browser suggests to prefill a field. This is not the case here as the field is secured to not allow that. Please allow this.

screenshot from 2018-08-23 12-03-07

djcruz93:
- Use autocomplete attribute

Fixes #2081

### DIFF
--- a/src/www/index.html
+++ b/src/www/index.html
@@ -304,7 +304,7 @@
           </li>
         </ul>
         
-        
+
         <!-- Map Integration-->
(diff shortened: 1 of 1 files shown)


iamareebjamal:
Please don't innovate PR titles, use the present guidelines

PR #2411, opened 2020-11-30
fix: Run a batch of 15 tasks in parallel

dhruvjain99:
#### Checklist

- Earlier since there was no limit on the batch size so due to limited resources any subsequent request got queued in the last of these tasks
- Now since we are processing in a batch of 15, any new request will only get queued at the end of these 15 tasks

- Used async.eachLimit with 15 as the limit of parallel task at a time instead of async.each which had no limit.


dhruvjain99:
@iamareebjamal can you please review this?

